Issue #11, opened 2019-07-16
satfunc2df only works for one SATNUM


eivindsm:
Function satfunc2df fails to parse tables with more than 1 satnum due to upstream issue in sunbeam. Testfunction conveniently passes due to the fact that it only has one SATNUM in test.

berland:
Is this a known or unknown sunbeam issue?

Should we resort to an existing Python-parser in the meantime?

berland:
Test that exposes bug: https://github.com/equinor/ecl2df/tree/satfuncbug

berland:
Trying to parse the swofstr without alll the band-aid in 
https://github.com/equinor/ecl2df/blob/cf1662bdafadc7944fe7d01b58f92a90b241fb98/ecl2df/eclfiles.py#L17
we get the error from sunbeam:
```
ValueError: PARSE_MISSING_DIMS_KEYWORD: Expected the kewyord: TABDIMS to infer the number of records in: SWOF
```
 Perhaps opm-common/sunbeam is unable to parse SWOF without knowing in advance the number of SATNUMs it contains?

berland:
Confirmed by trial and error that sunbeam requires TABDIMS with NTSFUN set correctly to parse SWOF.

TABDIMS with correct NTSFUN can be injected into the string if it is not found in the string/deck. There should be an option for the user to supply it when it is known.

It might be possible for satfunc2df, when TABDIMS is not found in the deck, to estimate it from
trial and error, progressing from 1 SATNUM and upwards (and not including (PARSE_RANDOM_SLASH, sunbeam.action.ignore))

berland:
Solved in #14

PR #14, opened 2019-08-26
Make it possible to parse SWOF/SGOF without TABDIMS present

berland:
TABDIMS/NTSFUN can now be inferred/guessed.

This is done by trial-and-error, determined by when sunbeam fails.


berland:
Solves #11

berland:
When TABDIMS is not present, number of saturation functions is inferred from 
automated trial-and-error using sunbeam parsing.

PR #27, opened 2019-09-03
Add feature to only output vertical nnc

berland:


berland:
Test should be smarter.

Code would be quicker if filtering takes place prior to building dataframe.

Not intuitive behaviour of in-place edits done by `filter_vertical()`.

Should be renamed to `filter_pillars()` or something? `-v` cannot be used as short-hand in arguments, as it is taken by `--verbose`

PR #29, opened 2019-09-06
PORV support in grid2df, inc. API change.

### LINKED ISSUE #10: Read PORV from INIT file
The PORV header in Eclipse INIT files is ignored by ecl2df since it has a different number of elements than the usual ones.

However, it just needs to be correctly index mapped in order to be included, check 
https://github.com/OPM/opm-output/blob/d5ac13049b15aecbacadbca7f3e92a2baa155ed3/opm/output/eclipse/EclipseIO.cpp#L249

berland:
Resolves #10


berland:
Solving #10

PR #84, opened 2020-01-07
Avoid AttributeError on py3 when ecl2csv is called with no args

berland:
This breaks the Py2 argparse API.


berland:
This only works with recent Argparse library versions.

berland:
This was a Py36 vs Py37 issue. Solved in #122

Issue #93, opened 2020-02-05
[grid] initkeys not full working with grid.df


berland:
vectors="PRESSURE" seems to give a 'no values to concatenate' error, even it it is in the dataframe returned when vectors="*"

berland:
Credit: hsum

PR #96, opened 2020-02-05
Fix bug with grid.df() when only rst vectors are asked for

berland:
Also extend the df() api with stackdates, and add tests


berland:
Resolves #93

PR #107, opened 2020-03-09
Update common.py

wouterjdb:
When reading a correct FAULTS keyword definition using ecl2df, with opm-common master branch clone from Mar 6, 2020 I bumped into the following error:

image

The change in this pull request fixes the problem. I'm unsure if this might lead to other unintended results. Please review.

### DIFF
--- a/ecl2df/common.py
+++ b/ecl2df/common.py
@@ -96,7 +96,10 @@ def parse_opmio_deckrecord(
 
     for item_idx, jsonitem in enumerate(itemlist):
         item_name = jsonitem["name"]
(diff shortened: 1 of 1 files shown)


wouterjdb:
So from the failing tests it clearly has some unwanted side-effects.....

berland:
Confirmed that test errors are due to upstream changes in opm-common. Use the tag testing/2020.3/rc1 for now.

To be resolved after opm-common.2020.3 is released

wouterjdb:
OK will do. Thanks.

berland:
Fixed in #111

Issue #113, opened 2020-03-19
[summary] converting to date index to datetime by default


miroine:
In the current behavior the indexes (Date) are an object ... would be much easier if they were by default as a datetime64 
![image](https://user-images.githubusercontent.com/40567543/77050133-8f324200-69c9-11ea-9363-ab9776806942.png)

berland:
It makes technically sense to return it as datetime64, but I am not convinced it will be easier for users in all cases. But this is not cast in stone, and if it is to be changed, it should not be changed to late.

asnyv:
For me it is not that important if it is a Date object or datetime64/Pandas timestamp, but would prefer it to be the same all the way. I see that fmu-ensemble (and I assume ecl2df as it is a copy of fmu-ensemble for summaries) returns datetime64 for "raw" time_index, and the Date object if you define a time_index like "monthly", which isn't that easy to guess before I stumble opon it. 
In general I would prefer most of the equivalents like `ecl.summary.EclSum.pandas_frame()`,` fmu-ensemble`'s` .get_smry()/.load_smry()` and `ecl2df.summary.df()` to return at as similar formats as possible

berland:
Any change in default behaviour of this must be coordinated with fmu-ensemble. Consequences for users must be examined upfront. A less intrusive change could be to add an option `alwaysdatetime=True` (default is False as now) to the API, so that caller functions can choose to always have the DATE column converted to Datetime when returned.

asnyv:
I could live with that @berland 👍

berland:
#134 is merged, so API users can always get this behaviour when wanted.

berland:
PR #207 is taking one step towards returning DatetimeIndex, by issuing a FutureWarning if an explicit `datetime=True` is not passed to `summary.df()`

berland:
Note: Pandas has a hard limit at 2262-04-11 when using DatetimeIndex (64-bits and nanosecond accuracy only allows representing 584 years). Eclipse simulations can go on for longer than year 2262.

berland:
Context: https://github.com/pandas-dev/pandas/issues/7307

berland:
The foreseeable future in reservoir simulation is defined as 500 years. This is a showstopper for having DatetimeIndex as default in Pandas dataframes.

PR #134, opened 2020-04-14
Let summary.df be able to always return datetime index

berland:
Suggestion from #113


tests/test_summary.py
```diff
@@ -23,11 +23,19 @@ def test_summary2df():
...
+    assert sumdf.index.name == "DATE"
+    assert sumdf.index.dtype == "datetime64[ns]" or sumdf.index.dtype == "datetime64"
+
     assert not sumdf.empty
     assert sumdf.index.name == "DATE"
     assert not sumdf.columns.empty
     assert "FOPT" in sumdf.columns
 
+    sumdf = summary.df(eclfiles, datetime=True)
+    # (datetime=True is superfluous when raw timereports are requested)
```
asnyv, line 35:
```suggestion
    # (datetime=True is superfluous when raw time reports are requested)
```

PR #161, opened 2020-06-17
opm-common from PyPI

anders-kiaer:


anders-kiaer:
Closing this as a test of the newly released opm-common binaries - stil waiting on official opm PyPI version.

PR #163, opened 2020-06-22
Use official opm from PyPI

anders-kiaer:


setup.py
```diff
@@ -23,6 +23,7 @@
 TEST_REQUIREMENTS = [
     "black; python_version >= '3'",
     "networkx",
+    "opm; python_version >= '3.5'",
```
berland, line 26: Move to REQUIREMENTS?

anders-kiaer: Heh, yes. Good catch.

PR #177, opened 2020-09-16
Add ecl2csv and csv2ecl as ERT forward models

berland:
Add CSV2ECL and ECL2CSV as ERT  forward models.

No options are possible  through these, only DATAFILE and output file. Allowing options probably require dedicated forward models for each ecl2csv "subcommand".

### DIFF
--- a/ecl2df/hook_implementations/jobs.py
+++ b/ecl2df/hook_implementations/jobs.py
@@ -1,0 +1,58 @@
+import importlib
+import os
+from pkg_resources import resource_filename
+
+from ert_shared.plugins.plugin_manager import hook_implementation
+from ert_shared.plugins.plugin_response import plugin_response
+
+
+def _get_jobs_from_directory(directory):
+    resource_directory = resource_filename("ecl2df", directory)
+
+    all_files = [
+        os.path.join(resource_directory, f)
+        for f in os.listdir(resource_directory)
+        if os.path.isfile(os.path.join(resource_directory, f))
+    ]
+    return {os.path.basename(path): path for path in all_files}
+
+
+@hook_implementation
+@plugin_response(plugin_name="ecl2df")
+def installable_jobs():
+    return _get_jobs_from_directory("config_jobs")
+
+
+def _get_module_variable_if_exists(module_name, variable_name, default=""):
+    try:
+        script_module = importlib.import_module(module_name)
+    except ImportError:
+        return default
+
+    return getattr(script_module, variable_name, default)
+
+
+@hook_implementation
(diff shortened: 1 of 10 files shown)


ecl2df/hook_implementations/jobs.py
```diff
@@ -0,0 +1,64 @@
...
+@hook_implementation
+@plugin_response(plugin_name="ecl2df")
+def installable_jobs():
+    return _get_jobs_from_directory("config_jobs")
+
+
+@hook_implementation
+@plugin_response(plugin_name="ecl2df")
+def installable_workflow_jobs():
+    return {}
```
sondreso, line 29: This is not needed if it's not returning anything, but if this is needed in the future I'm fine with leaving it as is!

tests/test_ert_hooks.py
```diff
@@ -0,0 +1,86 @@
...
+
+import pytest
+
+try:
+    from ert_shared.plugins.plugin_manager import ErtPluginManager
+except ImportError:
+    pytest.skip(
+        "ERT is not installed, skipping hook implementation tests.",
+        allow_module_level=True,
+    )
```
sondreso, line 14: I'm a little bit worried that if the module structure of ERT changes, this test will just be skipped and everything appears green. Maybe do something like this instead?
```python
try:
    import ert_shared
except ImportError:
    pytest.skip(
        "ERT is not installed, skipping hook implementation tests.",
        allow_module_level=True,
    )

from ert_shared.plugins.plugin_manager import ErtPluginManager
```

tests/test_hook_implementations.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+try:
+    import ecl2df.hook_implementations.jobs
+    from ert_shared.plugins.plugin_manager import ErtPluginManager
+except ImportError:
+    pytest.skip(
+        "ERT is not installed, or Python2. Skipping hook implementations.",
+        allow_module_level=True,
+    )
```
sondreso, line 15: Same as the other file :slightly_smiling_face:

sondreso (approved):
Other than my two minor comments, this LGTM!

ecl2df/hook_implementations/jobs.py
berland: Removed.

tests/test_ert_hooks.py
berland: Fixed.

tests/test_hook_implementations.py
berland: Fixed.

PR #186, opened 2020-10-19
Add WELOPEN support in ecl2df COMPDAT parsing

wouterjdb:


ecl2df/compdat.py
```diff
@@ -241,6 +255,76 @@ def unrolldf(dframe, start_column="K1", end_column="K2"):
...
+                (compdat_df["WELL"] == row["WELL"])
+                & (compdat_df["DATE"] <= row["DATE"])
+                & (compdat_df["I"] == row["I"])
+                & (compdat_df["J"] == row["J"])
+                & (compdat_df["K1"] == row["K"])
+                & (compdat_df["K2"] == row["K"])
+            ].drop_duplicates(subset=["I", "J", "K1", "K2"], keep="last")
+        elif row["C1"] or row["C2"]:
+            raise ValueError(
+                "Lumped connections are not supported in the a WELOPEN keyword."
```
berland, line 293: Typo; "the a"

berland: Since this is a ValueError, it will err hard. Which will make it impossible to extract only COMPDAT from a random deck where C1 and C2 is used. Downgrade to WARNING and skip the current KeyWord (or catch the ValueError where `applywelopen()` is performed). (and if exception is to be used, I wonder if there is a NotImplementedError that can be thrown instead?)

ecl2df/compdat.py
```diff
@@ -241,6 +255,76 @@ def unrolldf(dframe, start_column="K1", end_column="K2"):
...
+            )
+        elif not (row["I"] and row["J"] and row["K"]):
+            previous_state = compdat_df[
+                (compdat_df["WELL"] == row["WELL"])
+                & (compdat_df["DATE"] <= row["DATE"])
+            ].drop_duplicates(subset=["I", "J", "K1", "K2"], keep="last")
+        else:
+            raise ValueError(
```
berland, line 301: This situation more warrants a hard error, so catching all ValueError's  will probably be to  broad?

ecl2df/compdat.py
```diff
@@ -241,6 +255,76 @@ def unrolldf(dframe, start_column="K1", end_column="K2"):
...
+            )
+
+        if previous_state.empty:
+            raise ValueError(
+                "A WELOPEN keyword is not acting on any existing connection. "
+                "(I=%s,J=%s,K=%s)" % (row["I"], row["J"], row["K"])
+            )
+
+        new_state = previous_state
+        new_state["OP/SH"] = row["STATUS"]
```
berland, line 313: Maybe go trough the `COMPDAT_RENAMER` for this one, in order to "document" why this is done.

tests/test_welopen.py
```diff
@@ -0,0 +1,269 @@
...
+        pd.DataFrame(
+            {
+                "WELL": {0: "OP1", 1: "OP1", 2: "OP1"},
+                "I": {0: 33, 1: 34, 2: 35},
+                "J": {0: 110, 1: 111, 2: 111},
+                "K1": {0: 31, 1: 32, 2: 33},
+                "K2": {0: 31, 1: 32, 2: 33},
+                "OP/SH": {0: "SHUT", 1: "OPEN", 2: "SHUT"},
+                "SATN": {0: 0, 1: 0, 2: 0},
+                "TRAN": {0: None, 1: None, 2: None},
```
berland, line 43: (could probably have added some `dropna` to the test-code to avoid listing these)

tests/test_welopen.py
```diff
@@ -0,0 +1,269 @@
+import datetime
+
+import pandas as pd
+
+WELOPEN_CASES = [
+    (
+        """
+    DATES
```
berland, line 8: The indent here is not Eclipse-syntax, but I presume it works?

tests/test_compdat.py
```diff
@@ -181,6 +182,53 @@ def test_tstep():
     assert "2001-05-07" in dates
 
 
+@pytest.mark.parametrize("test_input,expected", WELOPEN_CASES)
```
berland, line 185: I am not sure, but I would have put this entire function into `test_welopen.py`. It should be ok to split the testing of the compdat-module into more than one file (even though this is the first file doing that).

ecl2df/compdat.py
```diff
@@ -241,6 +255,76 @@ def unrolldf(dframe, start_column="K1", end_column="K2"):
...
+                (compdat_df["WELL"] == row["WELL"])
+                & (compdat_df["DATE"] <= row["DATE"])
+            ].drop_duplicates(subset=["I", "J", "K1", "K2"], keep="last")
+        else:
+            raise ValueError(
+                "A WELOPEN keyword contains data that could not be parsed. "
+                "(I=%s,J=%s,K=%s)" % (row["I"], row["J"], row["K"])
+            )
+
+        if previous_state.empty:
```
berland, line 306: A `pytest.raises` test would be fine on this, f.ex when WELOPEN is applied on a not-defined well.

ecl2df/compdat.py
```diff
@@ -241,6 +255,76 @@ def unrolldf(dframe, start_column="K1", end_column="K2"):
...
+                "Lumped connections are not supported in the a WELOPEN keyword."
+            )
+        elif not (row["I"] and row["J"] and row["K"]):
+            previous_state = compdat_df[
+                (compdat_df["WELL"] == row["WELL"])
+                & (compdat_df["DATE"] <= row["DATE"])
+            ].drop_duplicates(subset=["I", "J", "K1", "K2"], keep="last")
+        else:
+            raise ValueError(
+                "A WELOPEN keyword contains data that could not be parsed. "
```
berland, line 302: Is there any known cases where this would occur (could be nice to add a pytest.raises test)? Or is it totally bogus data that would trigger this?

ecl2df/compdat.py
wouterjdb: This is a value error as the combination of values don't make sense (I and J is given, but not K). Of course one could write a error catching for all individual values... But I guess this will do.

ecl2df/compdat.py
wouterjdb: As said, incomplete input. For a valid case it shouldn't end up here.

ecl2df/compdat.py
wouterjdb: fixed

ecl2df/compdat.py
wouterjdb: fixed, downgraded to warning using the logger

ecl2df/compdat.py
```diff
@@ -241,6 +255,76 @@ def unrolldf(dframe, start_column="K1", end_column="K2"):
     return unrolled
 
 
+def applywelopen(compdat_df, welopen_df):
+    """Apply WELLOPEN actions to the COMPDAT dataframe.
```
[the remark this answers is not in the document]
wouterjdb: fixed

ecl2df/compdat.py
wouterjdb: The `COMPDAT_RENAMER` will not fix this. Then I would have to make a new one as "`OP/SH`" is "`STATUS`" in `WELOPEN` and "`STATE`" in `COMPDAT`. Do you prefer that?

tests/test_compdat.py
wouterjdb: Move function. Kept "test_welopen" in here, as that tests directly a function in the compdat.py file.... not entirely sure though. What do you think?

tests/test_welopen.py
wouterjdb: fixed.

tests/test_welopen.py
wouterjdb: Yeah, it works. But there was some variation in the actual indent. All now have 1 space.

wouterjdb (commented):
Thanks 👍

ecl2df/compdat.py
berland: Ok. My intention of the comment was more to ensure it gets documented "why" this line (313) is there, because it looks odd when reading the code.

tests/test_compdat.py
berland: fine!

ecl2df/compdat.py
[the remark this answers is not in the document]
wouterjdb: I see know that I actually did not yet fix the second part of your comment. Will do that as well still.

ecl2df/compdat.py
wouterjdb: I will place a comment there.

ecl2df/compdat.py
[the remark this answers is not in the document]
wouterjdb: fixed

ecl2df/compdat.py
wouterjdb: fixed

berland:
One thing, can you add `WELOPEN` to the shell-script in `opmkeywords/runmetoupdate.sh`? And then rerunning the shell script manually should not make any changes (I think your submitted WELOPEN was identical)

berland (approved):
Good!

wouterjdb:
Done.

Seems like there were some changes to files though (not `WELOPEN`). That should maybe be addressed in some other PR.

berland:
> Done.
> 
> Seems like there were some changes to files though (not `WELOPEN`). That should maybe be addressed in some other PR.

Yes, leave them for later.

PR #189, opened 2020-10-29
Fix issues with WELOPEN and COMPDAT entries at the same date

wouterjdb:
`WELOPEN` previously would have overwritten all actions in at the current date; now, the order of the keywords inside of the date-block dictates what happens as it would in a simulation run.

The following SCHEDULE section,
```
DATES
 x x x /

WELOPEN
 OP1 SHUT /
/
COMPDAT
 OP1 1 1 1 OPEN /
/
```
now produces one `COMPDAT` line with `OPEN` instead of `SHUT` one.

### DIFF
--- a/ecl2df/compdat.py
+++ b/ecl2df/compdat.py
@@ -346,9 +346,7 @@ def applywelopen(compdat_df, welopen_df):
     if not compdat_df.empty:
         compdat_df = (
             compdat_df.sort_values(by=["KEYWORD_ID"])
-            .drop_duplicates(
-                subset=["I", "J", "K1", "K2", "DATE", "KEYWORD_ID"], keep="last"
-            )
+            .drop_duplicates(subset=["I", "J", "K1", "K2", "DATE"], keep="last")
             .reset_index(drop=True)
         )
 
--- a/ecl2df/compdat.py
+++ b/ecl2df/compdat.py
@@ -140,7 +140,6 @@ def deck2dfs(deck, start_date=None, unroll=True):
                 )
                 if "INFO_TYPE" in rec_data and rec_data["INFO_TYPE"] == "ABS":
                     rec_data["SEGMENT_MD"] = rec_data["SEGMENT_LENGTH"]
-
                 welsegsrecords.append(rec_data)
 
     compdat_df = pd.DataFrame(compdatrecords)
--- a/ecl2df/compdat.py
+++ b/ecl2df/compdat.py
@@ -93,7 +93,7 @@ def deck2dfs(deck, start_date=None, unroll=True):
                     rec, "COMPDAT", renamer=COMPDAT_RENAMER
                 )
                 rec_data["DATE"] = date
-                rec_data["KEYWORD_ID"] = idx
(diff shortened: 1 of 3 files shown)


ecl2df/compdat.py
```diff
@@ -93,6 +93,7 @@ def deck2dfs(deck, start_date=None, unroll=True):
                     rec, "COMPDAT", renamer=COMPDAT_RENAMER
                 )
                 rec_data["DATE"] = date
+                rec_data["KEYWORD_ID"] = idx
```
berland, line 96: `KEYWORD_ID` -> `KEYWORD_IDX` ?

ecl2df/compdat.py
```diff
@@ -155,6 +157,9 @@ def deck2dfs(deck, start_date=None, unroll=True):
     if unroll and not welsegs_df.empty:
         welsegs_df = unrolldf(welsegs_df, "SEGMENT1", "SEGMENT2")
 
+    if "KEYWORD_ID" in compdat_df.columns:
```
berland, line 160: Should we drop it, or could it become handy to keep for some other use later?

berland: (who knows what problems will creep  up when we finally manage to auto-merge these three dataframes)

berland: thinking of issue #6

wouterjdb: It will definitely come in handy - for the same reasons. But I'd say we can mention that in that issue (extend usage of `KEYWORD_IDX`). If I would not return `KEYWORD_IDX` I'd most likely break many other things (at least some tests).

berland:
Can we have a test where WELOPEN only changes a portion of a well?

ecl2df/compdat.py
berland: ok!

berland:
Can we have a test where WELOPEN is repeated, first SHUT and then OPEN (or reverse) in sequence, with no DATES or COMPDAT in between them? Probably valid E100, although it is strange  setup.

wouterjdb:
Added the tests 🧪

berland:
It looks good. My hunch is that there are more corner cases to cater for, for which tests are the only gold. Can you look a little bit for more sharp corners?

tests/test_welopen.py
```diff
@@ -331,6 +331,60 @@
...
+     1 MAY 2001 /
+    /
+
+    COMPDAT
+     'OP1' 1 1 1 1 'OPEN'  /
+    /
+
+    WELOPEN
+     'OP1' 'OPEN' /
+     'OP1' 'SHUT' /
```
berland, line 346: nice!

wouterjdb:
Added two more tests 🧪 on some corner cases which are most likely accepted in E100, but a in the 'user error' territory. That by the way showed that the previous usage of Python's build-in `all()`-method actually didn't do what I intended it to do. Now using Pandas `all()`-method on the dataframe. So, something good came out of it 👍

berland (approved):
LGTM

PR #193, opened 2020-11-04
Drop Py2. Only PyPi dependencies. OPM-2020-10.

berland:
Thrown exception has apparently changed. All tests pass with opm-common from git HEAD with this patch.


anders-kiaer (approved):
:rocket: :tada:

PR #206, opened 2020-11-18
Replace os.path with pathlib

berland:


ecl2df/common.py
```diff
@@ -55,9 +55,7 @@
     "WELSEGS",
     "WELSPECS",
 ]:
-    with open(
-        os.path.join(os.path.dirname(__file__), "opmkeywords", keyw), "r"
-    ) as file_handle:
+    with open(Path(__file__).parent / "opmkeywords" / keyw, "r") as file_handle:
         OPMKEYWORDS[keyw] = json.load(file_handle)
```
anders-kiaer, line 59: A direct alternative when using a `Path` object is:
```suggestion
        OPMKEYWORDS[keyw] = json.load((Path(__file__).parent / "opmkeywords" / keyw).read_text())
```

anders-kiaer (approved):

ecl2df/common.py
berland: Fixed! thx.

PR #207, opened 2020-11-19
Adapt to be used directly with fmu-ensemble

berland:
Based on what is needed to pass tests in fmu-ensemble

Necessary for able https://github.com/equinor/fmu-ensemble/issues/166

 * [x] Tests for the changes in ecl2df must be transferred from fmu-ensemble's test suite.
 * [x] Pylinting

Bonus changes:
* Deprecate non-datetime-index as return type from `df()`
* Refactor/deletion for code readability, normalize_dates() has been removed (the function is assumed to be private)

### DIFF
--- a/ecl2df/eclfiles.py
+++ b/ecl2df/eclfiles.py
@@ -146,7 +146,15 @@ class EclFiles(object):
                     errno.ENOENT, os.strerror(errno.ENOENT), smryfilename
                 )
             logger.info("Opening UNSMRY file: %s", smryfilename)
-            self._eclsum = EclSum(smryfilename, include_restart=include_restart)
+            try:
+                self._eclsum = EclSum(smryfilename, include_restart=include_restart)
+            except IOError:
+                # This can happen if there is something wrong with the file
+                # or if SMSPEC is missing.
+                logger.warning(
+                    "Failed to create summary instance from %s", smryfilename
+                )
+                return None
         return self._eclsum
 
     def get_initfile(self):
--- a/ecl2df/eclfiles.py
+++ b/ecl2df/eclfiles.py
@@ -146,15 +146,7 @@ class EclFiles(object):
                     errno.ENOENT, os.strerror(errno.ENOENT), smryfilename
                 )
             logger.info("Opening UNSMRY file: %s", smryfilename)
-            try:
-                self._eclsum = EclSum(smryfilename, include_restart=include_restart)
-            except IOError:
-                # This can happen if there is something wrong with the file
-                # or if SMSPEC is missing.
-                logger.warning(
-                    "Failed to create summary instance from %s", smryfilename
-                )
-                return None
+            self._eclsum = EclSum(smryfilename, include_restart=include_restart)
         return self._eclsum
 
     def get_initfile(self):
--- a/ecl2df/summary.py
+++ b/ecl2df/summary.py
@@ -17,11 +17,17 @@ from .common import write_dframe_stdout_file
 
 logger = logging.getLogger(__name__)
 
+# Frequency mnemonics for the API consumer to use:
+FREQ_RAW = "raw"
+FREQ_FIRST = "first"
+FREQ_LAST = "last"
 PD_FREQ_MNEMONICS = {
-    "monthly": "MS",
-    "yearly": "YS",
     "daily": "D",
     "weekly": "W-MON",
+    "monthly": "MS",
+    "yearly": "YS",
+    # Any frequency mnemonics not mentioned here will be
+    # passed on to Pandas.
 }
 """Mapping from ecl2df custom offset strings to Pandas DateOffset strings.
 See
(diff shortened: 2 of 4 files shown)


BryanLaura (approved):

ecl2df/eclfiles.py
```diff
@@ -146,7 +146,15 @@ def get_eclsum(self, include_restart=True):
...
                 )
             logger.info("Opening UNSMRY file: %s", smryfilename)
-            self._eclsum = EclSum(smryfilename, include_restart=include_restart)
+            try:
+                self._eclsum = EclSum(smryfilename, include_restart=include_restart)
+            except IOError:
+                # This can happen if there is something wrong with the file
+                # or if SMSPEC is missing.
+                logger.warning(
+                    "Failed to create summary instance from %s", smryfilename
```
jondequinor, line 155: Why is a failure a warning and not an actual failure?

ecl2df/summary.py
```diff
@@ -1,7 +1,11 @@
 """Provide a two-way Pandas DataFrame interface to Eclipse summary data (UNSMRY)"""
 import logging
-import datetime
 from pathlib import Path
+import warnings
+
+# The name 'datetime' is in use by a function argument:
+from datetime import date as datetime_date
+from datetime import datetime as datetime_datetime
```
jondequinor, line 8: `dt` seems conventional, might be better here?

```python
import datetime as dt
…
```
But definitely unfortunate that `datetime` is a parameter 🤣

ecl2df/summary.py
```diff
@@ -105,47 +133,31 @@ def resample_smry_dates(
...
-        return datetimes
-    if freq == "first":
-        return [min(eclsumsdates).date()]
-    if freq == "last":
-        return [max(eclsumsdates).date()]
+    try:
+        parseddate = dateutil.parser.isoparse(freq)
+        return [parseddate]
+    except ValueError:
+        pass
```
jondequinor, line 148: Is swallowing the error here intentional? Maybe add a comment about why.

If not, maybe just add a `logger.debug` such that it is easier to debug.

ecl2df/summary.py
```diff
@@ -235,22 +248,38 @@ def df(
...
         )
     else:
+        # Can be None.
         time_index_arg = time_index
 
+    if isinstance(time_index_arg, list):
+        if len(time_index_arg) < 6:
+            time_index_str = str(time_index_arg)
+        else:
+            time_index_str = str(time_index_arg[0:3] + ["..."] + time_index_arg[-3:])
```
jondequinor, line 258: This is for presentational purposes I assume? Note that `[...]` is special notation in python, so I think
```python
f"{time_index_arg[0:3]} … {time_index_arg[-3:]}"
```

Might be less confusing?

ecl2df/eclfiles.py
berland: Right, degrading the failure should not happen here, but in fmu.ensemble if it wants to. Changing to raise.. from

ecl2df/summary.py
berland: Changed to `dt`. Planning to remove the datetime parameter in the future.

ecl2df/summary.py
berland: added comments.

ecl2df/summary.py
berland: Suggestion incorporated!

ecl2df/summary.py
```diff
@@ -39,34 +42,56 @@ def date_range(start_date, end_date, freq):
...
+    only cropped or returned as is.
 
-    Args:
-        start_date: datetime.date
-        end_date: datetime.date
-        freq: string with either 'monthly' or 'yearly'.
-            Anything else will return the input as is
-    Return:
-        Tuple of normalized (start_date, end_date)
+    Arguments are the same as for resample_smry_dates
```
jondequinor, line 73: It doesn't hurt repeating this here.

ecl2df/summary.py
```diff
@@ -39,34 +42,56 @@ def date_range(start_date, end_date, freq):
...
     """
-    if freq in PD_FREQ_MNEMONICS:
-        freq = PD_FREQ_MNEMONICS[freq]
-    offset = pd.tseries.frequencies.to_offset(freq)
-    return (offset.rollback(start_date).date(), offset.rollforward(end_date).date())
+    if freq == "raw":
```
jondequinor, line 75: If parameters to `_drop_datelist` are the same as `resample_smry_dates` then these magic strings should really be shared between the two.

What exactly happens when `freq` is a date/datetime?

ecl2df/summary.py
```diff
@@ -39,34 +42,56 @@ def date_range(start_date, end_date, freq):
...
+            datetimes = [x for x in datetimes if x < end_date]
+            datetimes = datetimes + [end_date]
+        return datetimes
+    if freq == "first":
+        return [min(eclsumsdates).date()]
+    if freq == "last":
+        return [max(eclsumsdates).date()]
+    if isinstance(freq, (dt.date, dt.datetime)):
+        return [freq]
+    raise ValueError("BUG: Wrong arguments to _crop_datelist()")
```
jondequinor, line 94: Isn't it specifically the `freq` that is wrong here? Maybe raise 

```python
ValueError(f"freq argument must be this or that, not '{freq}'")
```

ecl2df/summary.py
```diff
@@ -86,7 +111,7 @@ def resample_smry_dates(
             Options for timeresampling are
             'daily', 'monthly' and 'yearly'.
             'last' will give out the last date (maximum),
```
jondequinor, line 113: Document `first` as well?

ecl2df/summary.py
jondequinor: Well, maybe it does. Ignore that comment.

ecl2df/summary.py
berland: magic strings replaced by module constants. 

if `freq` is a datetime, it is caught in the `if`-statement ~6 lines below.

ecl2df/summary.py
berland: Fixed.

ecl2df/summary.py
berland: Fixed.

jondequinor (approved):
Technically this looks good to me. Can't really comment on the functional. Maybe you want a second pair of eyes?

berland:
> Technically this looks good to me. Can't really comment on the functional. Maybe you want a second pair of eyes?

Thanks! I'll see if @asnyv  has any further comments before merge.

asnyv:
I think it looks ok @berland 
Doesn't seem like we have tweaked on existing behavior (except future warning)? And existing tests ok.

PR #208, opened 2020-11-20
Not configure root logger

anders-kiaer:


berland:
Testing `ecl2csv satfunc --verbose FOO.DATA` yields no output after this PR. (`satfunc` is just an example), so something more must be fixed.

berland:
(and this illustrates that there should be a `caplog` test in the test suite for testing the verbose option)

berland:
Adding `logging.basicConfig()` to `ecl2csv.py` and `csv2ecl.py` is perhaps the remedy?

anders-kiaer:
Yes, the application part (i.e. not the library itself) can set logging options. I can add something closer to the console entrypoints.

tests/test_logging.py
```diff
@@ -0,0 +1,28 @@
...
+        rstdates="",
+        dropconstants=False,
+        stackdates=False,
+        output=tmp_path / "eclgrid.csv",
+    )
+
+    grid_main(args)
+
+    if verbose:
+        assert caplog.records
```
berland, line 26: Looks good. You can also assert some strings to be present, like `assert "INFO" in caplog.text`

berland: (since any WARNING records could be in `caplog.records` irrespective of verbose (but that should also fail in the `else` below)

berland (approved):
Thx!

PR #214, opened 2020-12-02
Add support for writing synthetic UNSMRY files

berland:


ecl2df/summary.py
```diff
@@ -287,6 +287,72 @@ def df(
...
+      will be emitted for skipped columns
+
+    Args:
+        dframe (pd.DataFrame): Dataframe to read. Will not be modified.
+
+    Returns:
+        pd.DataFrame: Modified copy of incoming dataframe.
+    """
+    dframe = dframe.copy()
+    if "DATE" in dframe.columns:
```
karbor, line 305: Make this case insensitive? E.g allow "Date", "date" ...?

karbor: And move "Date" out of method, defined as a constant DATAFRAME_DATE_COLUMN_NAME="DATE" og deretter sjekke på feks `if DATAFRAME_DATE_COLUMN_NAME in [column.upper() for column in dframe.columns]` ?

ecl2df/summary.py
```diff
@@ -287,6 +287,72 @@ def df(
...
+        dframe["DATE"] = pd.to_datetime(dframe["DATE"])
+        dframe = dframe.set_index("DATE", drop=True)
+    dframe.sort_index(axis=0, inplace=True)
```
karbor, line 308: If "Date" was not in columns - check that the index is in fact of a date/datetime type and raise error if not?

karbor (approved):
Nice to get this feature supported. Comments are suggestions to make it more robust wrt how flexible it should be wrt user input. Depends on how strict we want to be in the CSV columns.

ecl2df/summary.py
berland: I am quite unsure to which degree we should support different variations of header names. All other columns are required to be exact on case and letters, and also for the other ecl2df modules.

ecl2df/summary.py
berland: Now done.

PR #216, opened 2020-12-04
Degrade ValueError in WELOPEN

berland:
When WELOPEN is attempted applied to non-existing connections,
skip these instead of erroring hard. Eclipse seems to just warn
in those cases


tklov11:
The problem is that ecl2df do not recognize that welopen is applied to a predefined list of wells.

wouterjdb (approved):
LGTM 👍

berland:
> The problem is that ecl2df do not recognize that welopen is applied to a predefined list of wells.

That problem is now covered by #218

berland:
Ok, based on #218 we should maybe hold this merge until it is confirmed that Eclipse actually will ignore WELOPEN on non-existing i,j,k.

tklov11 (approved):
Looks good to me.

berland:
Closing. Ecl2df should error hard when it thinks the output is wrong.

PR #220, opened 2020-12-09
Add numpy to dependencies

berland:
There are lines with `import numpy as np` in `ecl2df`, and it should thus be declared as a direct dependency, not only piggybacking pandas.


markusdregi (approved):
All the good stuff depends on `numpy` ;)

PR #224, opened 2020-12-15
Allow ecl2csv subcommand options in ERT forward model

berland:
This smells a little bit hacky:

* The job configuration must define a maximum number of options.
  Hopefully 5 options are enough
* The argparse setup had to be modified slightly, to cater for
  empty placeholders for the empty arguments strings that the forward
  model supplies


ecl2df/config_jobs/ECL2CSV
```diff
@@ -1,5 +1,16 @@
...
+DEFAULT <XARG3> ""
+DEFAULT <XARG4> ""
+DEFAULT <XARG5> ""
+DEFAULT <XARG6> ""
+DEFAULT <XARG7> ""
+DEFAULT <XARG8> ""
+DEFAULT <XARG9> ""
+DEFAULT <XARG10> ""
+
+ARGLIST <SUBCOMMAND> "--verbose" "--output" <OUTPUT> <XARG1> <XARG2> <XARG3> <XARG4> <XARG5> <XARG6> <XARG7> <XARG8> <XARG9> <XARG10> "--" <ECLBASE>
```
markusdregi, line 14: I don't mind 🤷 What we have done in libres forward models is to support `<OPTS>` defaulted to `""` and then you just pack all of the extra stuff into that one arguments...

berland: Hmm, maybe I forgot to try that..

berland: But does that work without a custom argparse?

berland: If I put say `"--time_index yearly"` in a forward model argument, ERT will complain:
`Error message: subst_list_add_from_string: something wrong with:<XARG1>=--time_index yearly  - spaces are not allowed in key or value part.`

berland: Pending review, I suggest merging this and injecting for komodo 2021.02.

sondreso (approved):
Think this looks fine, but don't know if @markusdregi has any more comments after his first review?

Issue #217, opened 2020-12-06
Missing support for WSEGVALV (ref #199)


tklov11:
For QC of WellBuilder with the Troll model, ecl2df should have support for the WSEGVALV keyword in addition to those mentioned in ref #199.

berland:
Closed by #254

PR #227, opened 2020-12-19
WSEGVALV support

tklov11:
Added functionality to read WSEGVALV from a data deck. Also added a test for the WSEGVALV keyword in test_compdat.py. This is in response to issue #217


berland:
Superseded by #228

PR #228, opened 2020-12-21
WSEGVALV support

tklov11:
Add functionality to read and test WSEGVALV keyword


berland:
The file `runmetoupdate.sh` should have WSEGVALV added.

tests/test_compdat.py
```diff
@@ -483,11 +483,17 @@ def test_msw_str2df():
 -- WELL   SEG  SEG2 ALPHA  SF             RHO     VIS  WCT
     OP_6  31   31   0.0001  -1.186915444  1000.0  1.0  0.5  /
 /
+
+WSEGVALV
+-- WELL   SEG CV             AC      L
+    OP_6  31       0.0084252 0.00075  1*  /
```
berland, line 489: This is the base-case test, proving the code works for "normal" input. But bugs are usually found in corner-cases with variable input, sometimes just stressing the implementation, sometimes the input is invalid. Can you add further tests with some more "exotic" input that the Eclipse manual says is valid? (I have not checked, but usually there is some fine print). What can be defaulted, and what cannot. Well-list will probably lead to trouble here as well.

tklov11: Yes, I have added three more tests: (1) An include-file from Reek, (2) Maximum default values and (3) WLIST. The WLIST option does not work so I have asserted versus a df of length 0, which gives a positive test until we have implemented WLIST in ecl2df. Then the test should be rewritten. Please let me know if this last test should be deleted.

tklov11: Had to fix a small typo (WLIST records end with '/') .

tklov11: Add WSEGAICD, WSEGSICD and WSEGVALV to runmetoupdate.sh + chmod uog-x opmkeywords/VSEGVALV

tests/test_compdat.py
```diff
@@ -500,3 +512,103 @@ def test_msw_str2df():
...
+/
+"""
+
+    # Test WSEGVALV max defaulted and wlist
+    deck = EclFiles.str2deck(schstr_valv_wlist)
+    compdfs = compdat.deck2dfs(deck)
+    wsegvalv = compdfs["WSEGVALV"]
+    assert len(wsegvalv) == 1
+    assert "WELL" in wsegvalv
+    assert wsegvalv["WELL"].unique()[0] != "OP_6"
```
berland, line 614: Add some sample tests towards the numerical values in the dataframe, checking that CV column actually has a specific value at the correct row perhaps?

berland:
The documentation on https://equinor.github.io/ecl2df/usage/compdat.html might deserve some update as well from this.

tklov11:
#228 is closed and #254 takes over.

PR #230, opened 2020-12-28
Avoid using opm.io private attributes

### LINKED ISSUE #184: Cleanup use of private methods in opm.common
See this line, consider if this is fixable either in opm-common 2020.10:

https://github.com/equinor/ecl2df/blame/master/ecl2df/common.py#L234

berland:
Resolves #184


wouterjdb (approved):
LGTM

Issue #210, opened 2020-11-26
ValueError: Keyword RTEMP is not properly terminated


tklov11:
Hi,

I try to access the compdat content using EclFiles as shown below:

f = ecl2df.Eclfiles('GINAKROG_PRED0_B5_WELL3_NEWPARAM.DATA')
c = ecl2df.compdat.df(f)

This gives the following error:

  File "<stdin>", line 1, in <module>
  File "/prog/res/komodo/2020.09.02-py36/root/lib/python3.6/site-packages/ecl2df/compdat.py", line 305, in df
    compdat_df = deck2dfs(eclfiles.get_ecldeck())["COMPDAT"]
  File "/prog/res/komodo/2020.09.02-py36/root/lib/python3.6/site-packages/ecl2df/eclfiles.py", line 97, in get_ecldeck
    deck = opm.io.Parser().parse(deckfile, parsecontext)
ValueError: Keyword RTEMP is not properly terminated

However, RTEMP looks properly terminated to me in the .DATA file. This error occurs in both the github and Komodo versions. The model runs fine in Eclipse.

Link:
.DATA file: /scratch/gina_krog/mpic/AICD_B5/eclipse-0/model/GINAKROG_PRED0_B5_WELL3_NEWPARAM.DATA

berland:
Thanks. Can you split into two issues?

tklov11:
Awh, do I really have to?

berland:
Do you foresee that one fix ( = one PR) will fix both problems? If not they should be split, so that each of then can be discussed and resolve/closed  separately. 

Just make a new issue for the second where you copy in the text from this issue, and then edit the this issue.

tklov11:
Ok then :-)

tklov11:
Does this look better?

berland:
I have edited the title, as the reference to Gina Krog does not matter for the bug.

Can you share details to make this reproducible without access to Gina Krog? Can you add RTEMP to the Reek example case in ecl2df f.ex and trigger the same?

You can copy the RTEMP statement into this issue.

tklov11:
Shure I can. I will start to reproduce RTEMP in Reek. I have to think about how to share details on the other issue, see if I can reproduce this also in Reek.

tklov11:
I included RTEMP and some arbitrary reservoir temperature in the reek model, see attached file. This is what I got.

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
[...]

PR #231, opened 2021-01-04
Add RTEMP to test deck

berland:
This requires OPM-common >= 2020.10 (>= 2020.10.2 on pypi)

Proves #210 is fixed.


asnyv (approved):
LGTM 👍

PR #235, opened 2021-01-06
Update JSON from opm, WCONHIST LIFT->ALQ

berland:
This incurs one API change for ecl2df, the column `LIFT` in the `wcon` dataframe output changes name to `ALQ`
~The `wcon` submodule already renames the  "LIFT" column to "ALQ", thus no API change.~

### DIFF
--- a/ecl2df/opmkeywords/SGFN
+++ b/ecl2df/opmkeywords/SGFN
@@ -7,6 +7,9 @@
     "keyword": "TABDIMS",
     "item": "NTSFUN"
   },
+  "requires": [
+    "GAS"
+  ],
   "items": [
     {
       "name": "DATA",
--- a/ecl2df/opmkeywords/SGOF
+++ b/ecl2df/opmkeywords/SGOF
@@ -7,6 +7,13 @@
     "keyword": "TABDIMS",
     "item": "NTSFUN"
   },
+  "requires": [
+    "GAS",
+    "OIL"
+  ],
+  "prohibits": [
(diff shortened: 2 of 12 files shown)


lars-petter-hauge:
Is the reason for updating that opm-common has been updated?

It also looks like there are more than a single entry that could break the api, but they are perhaps not reflected in the api?

berland:
Yes, these are "upstream"  API changes. Not sure whether API compatiblity for ecl2df should encompass column header names, but will be ok according to semantic versioning (less than 1.0 still).

For the SEG1 vs SEGMENT1 change, `compdat.py` is robust towards this change.

lars-petter-hauge:
I guess the json files here has to be a one-to-one match with the ones here: https://github.com/OPM/opm-common/tree/master/src/opm/parser/eclipse/share/keywords/000_Eclipse100? Could I ask the reason for why they have to be added here as well?

berland:
They should be one-to-one yes, but are duplicated for the sake of ensuring control over changes in column names.

tklov11 (approved):
LGTM. Checked additions against https://github.com/OPM/opm-common/blob/master/src/opm/parser/eclipse/share/keywords/000_Eclipse100/

Also checked full keyword definitions for WSEGx keywords.

berland:
There has been historically been bandage in `wcon.py` to circumvent column name changes, and there is some history to `WCONHIST` in opm.  Suggest to keep this as an API change to be in sync.

PR #236, opened 2021-01-06
Refactor pretty-printing of trees.

berland:
Incurs backwards compatible API change
 
Improved testing of tree stringification

Bugfix in edge_dataframe2dict on multiple roots.

Kudos to lgtm.com for exposing the problem in an alert.


tests/test_gruptree.py
```diff
@@ -272,8 +380,67 @@ def test_main(tmpdir):
 
 def test_prettyprint():
     """Test pretty printing via command line interface"""
-    sys.argv = ["ecl2csv", "gruptree", DATAFILE, "--prettyprint"]
-    ecl2csv.main()
+    commands = ["ecl2csv", "gruptree", DATAFILE, "--prettyprint"]
+    results = subprocess.run(commands, check=True, stdout=subprocess.PIPE)
```
lars-petter-hauge, line 384: I don't think it is wise to run tests in subprocess, it just adds an additional layer making debugging more difficult. Previous implementation modified `sys.argv`, which is also not a very nice thing to do, so I understand the reason to change. I would suggest that instead the test mocks `sys.argv` before calling `main`

berland: Ok, unaware of that (but I have a hunch someone actually recommended `subprocess.run` in tests elsewhere), but have been surprised recently that the coverage reports seemed to be able to see through `subprocess.run` .

lars-petter-hauge: Well, I would I least recommend mocking sys argv instead. Modifying it during runtime is worse, as you are actually modifying it for every test (so if you were to append anything to it in every test, you would end up with a very long sys.argv..). Sometimes you probably want to use subprocess, for instance if you want to verify that the entry point is correctly installed. I would reserve those tests to the simple scenario of `subprocess.run(["name", -h"])`

berland: Ok, thanks. Will use the opportunity to try `pytest-mock`

lars-petter-hauge:
I would really like to check it out locally and go through this PR in a debugger, would perhaps improve my questions a bit. I need to figure out if I can build opm locally instead, as it doesn't contain an osx release on pypi.

tests/test_gruptree.py
```diff
@@ -144,8 +210,8 @@ def test_grupnetdf():
                 ]
             ),
             """
-└── FIELD
-    └── OP
+FIELD
+└── OP
```
lars-petter-hauge, line 214: So what you want to improve in this PR is that you want to treat the top level differently? Are there more other intentions?

berland: Yes, the previous handling of the top-level is regarded as a bug.

The main intention of the PR was to solve a lgtm-issue, which triggered a rewrite to facilitate better testing, which again exposed the buggy top-level.

tests/test_gruptree.py
berland: Rebased, and a new commit added changing to mocking.

tests/test_gruptree.py
```diff
@@ -217,22 +281,61 @@ def test_grupnetroot(schstr, expected_dframe, expected_tree):
...
+            {"CHILD": "FIELDA", "PARENT": None},
+            {"CHILD": "FIELDB", "PARENT": None},
+            {"CHILD": "PLATA", "PARENT": "FIELDA"},
+            {"CHILD": "PLATB", "PARENT": "FIELDB"},
+        ]
+    )
+    list_of_treedicts = gruptree.edge_dataframe2dict(edges)
+    assert len(list_of_treedicts) == 2
+    assert {"FIELDA": {"PLATA": {}}} in list_of_treedicts
+    assert {"FIELDB": {"PLATB": {}}} in list_of_treedicts
```
lars-petter-hauge, line 305: Could you not simply assert the `list_of_treedicts` in a single go? Or does the `gruptree.edge_dataframe2dict` not guarantee order? Would also be possible to make it more obvious that the same result is expected twice - and reduce overhead.

tests/test_gruptree.py
```diff
@@ -123,6 +139,54 @@ def test_grupnetdf():
...
+foo
+├── bar
+└── com
+    └── fjooo""",
+        ),
+        # Integers as node names:
+        ({1: {2: {}}}, "1\n└── 2"),
+        # More complex structure:
+        # Note: Node names cannot be duplicated, even on
+        # unique branches:
```
lars-petter-hauge, line 164: Where does this restriction stem from?

tests/test_gruptree.py
lars-petter-hauge: Ah, right.

tests/test_gruptree.py
berland: It is from how `treelib` is used. I think the restriction can be relaxed within `ecl2df`, but have not tested.

tests/test_gruptree.py
berland: Fixed.

ecl2df/gruptree.py
```diff
@@ -225,16 +226,24 @@ def edge_dataframe2dict(dframe):
...
+    tree.create_node(name, name)
+    child_names = list(nested_dict.keys())
+    child_names.sort()
+    for child_name in child_names:
+        # The paste() function requires unique node names.
+        tree.paste(name, _subdict2treelib(nested_dict[child_name], child_name))
+    return tree
 
+
+def dict2treelib(nested_dict, deprecated=None):
```
lars-petter-hauge, line 245: A bit surprised that `treelib` didn't have a `treelib.from_dict` or similar (or just accepted a dict on construct), as it is such a general operation. But I see on their github page that several people have commented this, and although there's some work ongoing, they've yet to finalise it. 

I'm not sure if the best way to remove the first positional argument is to add another keyword argument (that requires some context in order to understand it's existence. It's really a private keyword argument to handle the case `dict2treelib("no longer possible", {})`and forward to `dict2treelib({})`). And if someone calls `dict2treelib(name="no longer possible", nested_dict={})` it'll also raise an error. Just not sure if it is worth it.. I don't have a good suggestion yet though.

Does it need to be a nested_dict by the way? Could it not just simpy be single entry dict? Would be handled I guess?

tests/test_gruptree.py
lars-petter-hauge: Would it invalidate data that otherwise is valid from the users perspective (as we are looking at representations of `GRUPNET` structures).

tests/test_gruptree.py
berland: Fixed once more to guarantee sorting by root-node-name.

tests/test_gruptree.py
berland: `GRUPTREE` (standard network model) I am quite sure do not allow repeated nodes (will give loops and recycled production(?)). Not that certain for `GRUPNET` (extended network model), the manual states that the tree must have "gathering tree topology", whatevery they mean by that. Extended network model is not yet supported by ecl2df.

ecl2df/gruptree.py
berland: I intend to remove this deprecated feature, but the backwards compatibility was added as the `rft.py` file actually used it, and then possibly other scripts. Correct that all-named arguments will now fail.

berland:
Superseded by #253

PR #238, opened 2021-01-11
Allow all subtools to output to stdout

### LINKED ISSUE #237: MAGIC_STDOUT, either all or none of subcommands should support dumping to stdout
Some subtools can have the output set to `-` as a command line option, and then ecl2csv will write to stdout instead of to file.

Make this consistent for all tools.

berland:
Resolves #237


eivindsm (approved):
LGTM

PR #251, opened 2021-01-22
Fix erroneous PVTG support in ecl2df.pvt

berland:
Apparently forgotten initially. Mistake was uncovered through adding test code warranted by codecov.io

### DIFF
--- a/ecl2df/pvt.py
+++ b/ecl2df/pvt.py
@@ -418,9 +418,7 @@ def df2ecl_pvtw(dframe, comment=None):
 
 
 def df2ecl_pvtg(dframe, comment=None):
-    """Print DENSITY keyword with data
-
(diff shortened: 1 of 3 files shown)


eivindsm (commented):
Looks like export of DENSITY is no longer supported; is it supported by another function in ecl2df ?

berland:
> Looks like export of DENSITY is no longer supported; is it supported by another function in ecl2df ?

It's still there in `df2ecl_density`, but `df2ecl_pvtg` will no longer erroneously dump density data..

berland:
> Looks like export of DENSITY is no longer supported; is it supported by another function in ecl2df ?

It's still there in `df2ecl_density`, but `df2ecl_pvtg` will no longer erroneously dump density data..

eivindsm (approved):
Perfect!

PR #252, opened 2021-01-22
Refactor include file disk dumping

berland:
Improve tests on the way

### DIFF
--- a/ecl2df/common.py
+++ b/ecl2df/common.py
@@ -1,6 +1,4 @@
-"""
-Common functions for ecl2df modules
-"""
+"""Common functions for ecl2df modules"""
 
 import sys
 import json
@@ -66,6 +64,9 @@ for keyw in [
     )
 
 
+# This is a magic filename that means read/write from/to stdout:
+__MAGIC_STDOUT__ = "-"
+
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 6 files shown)


ecl2df/common.py
```diff
@@ -84,7 +85,7 @@ def write_dframe_stdout_file(
         caller_logger (logging): Used if not stdout
         logstr (str): Logged if not stdout.
     """
-    if output == "-":
+    if output == __MAGIC_STDOUT__:
```
jondequinor, line 88: I think `None` would work just as well here? `-` usually is `stdin` in bash, might cause some confusion?

ecl2df/common.py
```diff
@@ -66,6 +64,9 @@
     )
 
 
+# This is a magic filename that means read/write from/to stdout:
+__MAGIC_STDOUT__ = "-"
```
jondequinor, line 68: Why the double dunder? Does this mean I cannot do this?

```python
from ecl2df.common import __MAGIC_STDOUT__

…
write_dframe_stdout_file(my_frame, __MAGIC_STDOUT__)
```

If that does not work, this is unfortunate.

ecl2df/common.py
jondequinor: Also, isn't `-` a perfectly valid file name?

jondequinor (changes requested):
Strongly recommend changing the `write_dframe_stdout_file` API.

ecl2df/common.py
berland: It works, but it creates a lot of problems. Many bash tools cannot create the file, `touch -` will not work. `echo foo > -` works, but `cat -` will not work afterwards. Not even `cat -- -`. You can also try `mkdir -` but it is hard to `cd -` afterwards.

ecl2df/common.py
berland: yes, that ddunder is probably misplaced.

PR #253, opened 2021-01-25
Refactor pretty-printing of trees without breaking API

berland:
Replaces #236, changing function name, and keeping old function with a ~DeprecationWarning~ FutureWarning.

Improved testing of tree stringification

Bugfix in edge_dataframe2dict on multiple roots.

Kudos to lgtm.com for exposing the problem in an alert.

### DIFF
--- a/ecl2df/gruptree.py
+++ b/ecl2df/gruptree.py
@@ -1,13 +1,10 @@
-#!/usr/bin/env python
-"""
-Extract GRUPTREE information from an Eclipse deck
-
-"""
+"""Extract GRUPTREE information from an Eclipse deck"""
 
 import sys
 import logging
 import datetime
 import collections
+import warnings
 
 import treelib
 import pandas as pd
@@ -207,8 +204,9 @@ def edge_dataframe2dict(dframe):
     Leaf nodes have empty dictionaries as their value.
 
     Returns:
-        list of nested dictionaries, as we in general
-            may have more than one root.
+        List of nested dictionaries, as we in general
+        may have more than one root. The list sorted
+        alphabetically on the top level node name.
     """
     if dframe.empty:
         return [{}]
@@ -225,10 +223,44 @@ def edge_dataframe2dict(dframe):
 
     children, parents = zip(*edges)
     roots = set(parents).difference(children)
-    trees = []
-    for root in list(roots):
-        trees.append({root: subtrees[root] for root in roots})
-    return trees
+    return [{root: subtrees[root]} for root in sorted(roots)]
+
+
+def _sub_tree_from_dict(nested_dict, name):
+    """Internal recursive function for generating trees."""
+    assert isinstance(nested_dict, dict)
+    tree = treelib.Tree()
+    tree.create_node(name, name)
+    child_names = list(nested_dict.keys())
+    child_names.sort()
+    for child_name in child_names:
+        # The paste() function requires unique node names.
+        tree.paste(name, _sub_tree_from_dict(nested_dict[child_name], child_name))
+    return tree
+
+
+def tree_from_dict(nested_dict):
+    """Convert a dictonary to a treelib Tree
+
+    The treelib representation of the trees is used
+    for pretty-printing (in ASCII) of the tree, you
+    only need to do str() of the returned result
+    from this function.
+
+    See `https://treelib.readthedocs.io/`
+
+    Args:
+        nested_dict: nested dictonary representing a tree.
+
+    Returns:
+        treelib.Tree
+    """
+    if not nested_dict:
+        # Return an empty string, because str(treelib.Tree()) will error
+        return ""
+    assert len(nested_dict.keys()) == 1, "One tree at a time please"
+    root_name = list(nested_dict.keys())[0]
+    return _sub_tree_from_dict(nested_dict[root_name], root_name)
 
 
 def dict2treelib(name, nested_dict):
@@ -249,6 +281,11 @@ def dict2treelib(name, nested_dict):
     Return:
         treelib.Tree
     """
+    warnings.warn(
+        "dict2treelib() is deprecated and will be removed, use tree_from_dict()",
+        FutureWarning,
+    )
+
     tree = treelib.Tree()
     tree.create_node(name, name)
     for child in nested_dict.keys():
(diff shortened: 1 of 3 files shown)


ecl2df/gruptree.py
```diff
@@ -1,13 +1,10 @@
-#!/usr/bin/env python
```
tklov11, line 1: Is this line needed?

berland: No, that is why it has been removed.

ecl2df/gruptree.py
```diff
@@ -1,13 +1,10 @@
...
-"""
-Extract GRUPTREE information from an Eclipse deck
-
-"""
+"""Extract GRUPTREE information from an Eclipse deck"""
```
tklov11, line 1: I liked the previous format better (without the empty line). Have used it for all docstrings in subscript-internal. Should I change?

ecl2df/gruptree.py
```diff
@@ -225,10 +223,44 @@ def edge_dataframe2dict(dframe):
...
     roots = set(parents).difference(children)
-    trees = []
-    for root in list(roots):
-        trees.append({root: subtrees[root] for root in roots})
-    return trees
+    return [{root: subtrees[root]} for root in sorted(roots)]
+
+
+def _sub_tree_from_dict(nested_dict, name):
+    """Internal recursive function for generating trees."""
```
tklov11, line 230: Inconsistent use of period at end of docstring.

ecl2df/gruptree.py
berland: It is the previous format that has the extra empty line, not sure if I understand the comment.

ecl2df/gruptree.py
berland: Yes, PEP-257 says we should end docstring sentences with a period. So it is the other functions that need a cleanup then.

Private (hidden) functions starting with a `_` are a blessing when you later want to change the API.. :)

berland:
(This PR also apparently fixes an unreported bug where multiple roots are displayed as duplicates of the first root)

ecl2df/gruptree.py
```diff
@@ -225,10 +223,44 @@ def edge_dataframe2dict(dframe):
...
     children, parents = zip(*edges)
     roots = set(parents).difference(children)
-    trees = []
-    for root in list(roots):
-        trees.append({root: subtrees[root] for root in roots})
-    return trees
+    return [{root: subtrees[root]} for root in sorted(roots)]
+
+
+def _sub_tree_from_dict(nested_dict, name):
```
lars-petter-hauge, line 229: This function creates a new tree instance for each child and pastes it onto the parent. Although this works, another possible approach is to follow along as the treelib documentation shows:

```
>>> tree = Tree()
>>> tree.create_node("Harry", "harry")  # root node
>>> tree.create_node("Jane", "jane", parent="harry")
>>> tree.create_node("Bill", "bill", parent="harry")
>>> tree.create_node("Diane", "diane", parent="jane")
```

If you go for a top-down recursive function instead, only a single tree is created:

```
def _add_to_tree_from_dict(nested_dict, name, tree, parent=None):
    assert isinstance(nested_dict, dict)

    tree.create_node(name, name, parent=parent)
    for key, value in sorted(nested_dict.items()):
        _add_to_tree_from_dict(nested_dict=value, name=key, tree=tree, parent=name)

def tree_from_dict(nested_dict):
    ....
    root_name = list(nested_dict.keys())[0]
    tree = treelib.Tree()
    _add_to_tree_from_dict(nested_dict[root_name], root_name, tree)
    return tree
```

There are different opinions about top-down recursive functions, and having to drag along an instance might not be the most suitable option. But I think I would favour only creating the tree once over the downside of having to pass it down.  I also think the code reads a bit better with this approach.

ecl2df/gruptree.py
```diff
@@ -225,10 +223,44 @@ def edge_dataframe2dict(dframe):
...
+    Args:
+        nested_dict: nested dictonary representing a tree.
+
+    Returns:
+        treelib.Tree
+    """
+    if not nested_dict:
+        # Return an empty string, because str(treelib.Tree()) will error
+        return ""
+    assert len(nested_dict.keys()) == 1, "One tree at a time please"
```
lars-petter-hauge, line 261: Should we provide a bit more feedback to the user? It might not be obvious what "one tree at a time" means

ecl2df/gruptree.py
```diff
@@ -225,10 +223,44 @@ def edge_dataframe2dict(dframe):
...
+
+    See `https://treelib.readthedocs.io/`
+
+    Args:
+        nested_dict: nested dictonary representing a tree.
+
+    Returns:
+        treelib.Tree
+    """
+    if not nested_dict:
```
lars-petter-hauge, line 258: The truthyness of python would allow this function to successfully run, although we might consider that it should fail.

Here we're allowing e.g. `tree_from_dict([])`, `tree_from_dict(0)` and `tree_from_dict(None)` to all give an empty string, although we explicitly say the argument should be a dictionary.

It's a bit nitpicky, but I figured the `if not nested_dict` was covering a bit much.

tests/test_gruptree.py
```diff
@@ -217,22 +287,68 @@ def test_grupnetroot(schstr, expected_dframe, expected_tree):
...
-    treelibtree = gruptree.dict2treelib("", gruptreedict[0])
-    treestring = str(treelibtree)
-    assert not treestring.strip()  # Let it return whitespace
+
+    treelibtree = gruptree.tree_from_dict(gruptreedict[0])
+    # Returning an empty string and not a treelib.Tree() is
+    # a workaround for a limitation in treelib.
+    assert treelibtree == ""
+
+    tmpdir.chdir()
```
lars-petter-hauge, line 347: Are these two different tests?

lars-petter-hauge:
Had a couple comments there in the end, but other than that I think the PR is shaping out :)

ecl2df/gruptree.py
berland: Fixed.

tests/test_gruptree.py
berland: Split.

ecl2df/gruptree.py
berland: Thanks, agree this looks better.

lars-petter-hauge:
I don't think that's the same lpha - I was a bit surprised as well, didn't know about it 😂

lars-petter-hauge (approved):
From what I can see all comments have been adressed. Looks good!

PR #254, opened 2021-01-25
Wsegvalv2

tklov11:
Oppdatert med støtte for WSEGVALV-nøkkelordet, tester og docs.


ecl2df/compdat.py
```diff
@@ -134,6 +135,12 @@ def deck2dfs(deck, start_date=None, unroll=True):
                 rec_data["DATE"] = date
                 rec_data["KEYWORD_IDX"] = idx
                 wsegaicdrecords.append(rec_data)
+        elif kword.name == "WSEGVALV":
+            for rec in kword:  # Loop over the lines inside WSEGAICD record
```
berland, line 139: Comment here looks wrong..

tklov11: yes, it does. I'll fix.

tklov11:
Do you have recommendations on other keywords that are tested vigorously, that I may have a look at?

berland (approved):
LGTM!

PR #257, opened 2021-01-26
Augment .pre-commit

berland:


.pre-commit-config.yaml
```diff
@@ -2,6 +2,8 @@ repos:
 - repo: https://github.com/pre-commit/pre-commit-hooks
   rev: v2.5.0
   hooks:
+  - id:  no-commit-to-branch
```
tklov11, line 5: If I understood correctly, the purpose of pre-commit is to give the contributor a chance to check the code in one simple command before comitting it from the working directory to his/her local repo. From this pre-commit yaml file I only see reference to flake8 and rstcheck. Should `black --check .` and `pytest tests/` also be included? These are the four tests I usually run in separate commands before I commit anything to my local repo.

berland: Pre-commit (after a local user has run `pre-commit install` in his/hers local copy) will add a hook to `git commit` so that these commands are checked every time a `git commit` is attempted.

Black is already included, but pytest should be excluded as it generally takes too long time. pre-commit should finish in a couple of seconds.

tklov11: ok

tklov11 (approved):
LGTM

PR #267, opened 2021-02-11
Revert deprecation on datetime indices, support 500 year timespan

berland:
Uncovering current bugs both in libecl and Pandas, circumvented
in this patch.

Pandas datetime indices are datetime64[ns] only, and is limited to year 2262, which
is not acceptable for ecl2df.

### DIFF
--- a/ecl2df/summary.py
+++ b/ecl2df/summary.py
@@ -221,7 +221,7 @@ def df(
     support for string mnenomics for the time index.
 
     The dataframe is always indexed by DATE, and the datatype for the
-    index will be usually be datetime64[ns] as long as all dates are
+    index will usually be datetime64[ns] as long as all dates are
     before year 2262. If a longer time range is detected, the index.dtype
     will be object, and consisting of datetime.datetime() objects. The index
     is always named "DATE".
@@ -425,7 +425,7 @@ def _fix_dframe_for_libecl(dframe: pd.DataFrame) -> pd.DataFrame:
                 index=dframe.index,
             )
 
-        dframe.set_index(dframe["DATE"], inplace=True)
+        dframe.set_index("DATE", inplace=True)
     if not isinstance(
         dframe.index.values[0], (dt.datetime, np.datetime64, pd.Timestamp)
     ):
@@ -450,7 +450,7 @@ def _fix_dframe_for_libecl(dframe: pd.DataFrame) -> pd.DataFrame:
             str({colname.partition(":")[0] + ":*" for colname in block_columns}),
         )
 
-    return dframe.drop("DATE", axis="columns", errors="ignore")
+    return dframe
 
 
 def df2eclsum(
--- a/ecl2df/summary.py
+++ b/ecl2df/summary.py
@@ -1,15 +1,16 @@
 """Provide a two-way Pandas DataFrame interface to Eclipse summary data (UNSMRY)"""
 import logging
 from pathlib import Path
-import warnings
 
 # The name 'datetime' is in use by a function argument:
 import datetime as dt
 
 import dateutil.parser
+import numpy as np
 import pandas as pd
 
-from ecl.summary import EclSum
+import ctypes
+from ecl.summary import EclSum, EclSumKeyWordVector
 
 from .eclfiles import EclFiles
 from . import parameters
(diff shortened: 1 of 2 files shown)


berland:
The injected and modified functions from libecl will be removed as soon as possible.

ecl2df/summary.py
```diff
@@ -188,7 +189,7 @@ def resample_smry_dates(
 
     datetimes = date_range(start_date_range, end_date_range, freq)
 
-    # Convert from Pandas' datetime64 to datetime.date:
+    # Convert from numpys datetime64 to datetime.date:
```
asnyv, line 192:
```suggestion
    # Convert from numpy's datetime64 to datetime.date:
```

ecl2df/summary.py
```diff
@@ -219,6 +220,12 @@ def df(
     This is a thin wrapper for EclSum.pandas_frame, by adding
     support for string mnenomics for the time index.
 
+    The dataframe is always indexed by DATE, and the datatype for the
+    index will be usually be datetime64[ns] as long as all dates are
```
asnyv, line 224:
```suggestion
    index will usually be datetime64[ns] as long as all dates are
```

ecl2df/summary.py
```diff
@@ -219,6 +220,12 @@ def df(
...
+    before year 2262. If a longer time range is detected, the index.dtype
+    will be object, and consisting of datetime.datetime() objects. The index
+    is always named "DATE".
```
asnyv, line 227: Not possible to always return `object` (`datetime.datetime`)? `pandas` living its own life? Not a fan of varying output types... Better to handle it here than in every dependecy?

ecl2df/summary.py
```diff
@@ -295,10 +302,9 @@ def df(
         # Warning is already logged by eclfiles.
         return pd.DataFrame()
 
-    dframe = eclsum.pandas_frame(time_index_arg, column_keys)
+    # dframe = eclsum.pandas_frame(time_index_arg, column_keys)
```
asnyv, line 305: Remove?
```suggestion
```

ecl2df/summary.py
```diff
@@ -427,7 +450,7 @@ def _fix_dframe_for_libecl(dframe: pd.DataFrame) -> pd.DataFrame:
             str({colname.partition(":")[0] + ":*" for colname in block_columns}),
         )
 
-    return dframe
+    return dframe.drop("DATE", axis="columns", errors="ignore")
```
asnyv, line 453: aren't any potential `DATE` column already dropped when in the line:
`dframe.set_index(dframe["DATE"], inplace=True)`? which will run if `DATE` is in `dframe.columns`?

ecl2df/summary.py
```diff
@@ -455,7 +478,90 @@ def df2eclsum(
         raise ValueError(f"Do not use dots in casename {casename}")
 
     dframe = _fix_dframe_for_libecl(dframe)
-    return EclSum.from_pandas(casename, dframe)
+    return _libecl_eclsum_from_pandas(casename, dframe)
+    # return EclSum.from_pandas(casename, dframe)
```
asnyv, line 482: remove?
```suggestion
```

ecl2df/summary.py
berland: I think we must store it as a string if we want to guarantee `object` all the time, and this is not tempting to enforce. Pandas is very eager about its DatetimeIndex.

asnyv: hmm, ok. Then I'm tempted to ask (though this is quite breaking, so probably not for this PR): Can we not just have `DATE` as a regular column and not care about the index at all? 😅

ecl2df/summary.py
berland: Only when I fix the `set_index(dframe["DATE"])` to `set_index('DATE')`. Fixed.

ecl2df/summary.py
berland: I kept these to signify that I want to roll back to using libecl for this.

ecl2df/summary.py
berland: I would prefer to always have `datetime.datetime()` in the index, and to Pandas, that is an index of type `object`. But pandas operations (like `apply()`) on indices with datetime.datetime jump to a DatetimeIndex when you poke it, as long as you are before year 2262. So I _think_ this variable output is something the API user must live with.

asnyv: Ok, I'll resign 😣🏳😅

ecl2df/summary.py
berland: fixed

ecl2df/summary.py
berland: fixed

ecl2df/summary.py
berland: Kept

asnyv (approved):
LGTM 😊

PR #268, opened 2021-02-15
Make ecl2df partially work on MacOS

berland:


berland:
@miroine, are you able to test?

miroine:
@berland ... working like a charm now in `mac `... thanks .... however u have to remove the `opm` dependency from the `setup.py `as well --> it s listed as a requirement there as well now

berland:
> @berland ... working like a charm now in `mac `... thanks .... however u have to remove the `opm` dependency from the `setup.py `as well --> it s listed as a requirement there as well now

It is intentionally listed there yes, that is why the documentation changes in this PR says you must do that manually. It can't  be removed from `setup.py`, it  would break installation for Linux users.

lars-petter-hauge:
I don't get it, is e.g. `guess_dim` never used? How would the setting `opm.io.action.ignore` not throw an `opm is not defined`?

I would have expected much more case checking behaviour, depending on how much of `opm.io` is used.

berland:
`ecl2df`  states `opm` as required, so the user experience for users not having opm.io available does not have high priority. This PR is a middle ground, to let it fail with whatever exception only when `opm` is really required.

lars-petter-hauge:
>ecl2df states opm as required, so the user experience for users not having opm available does not have high priority.

But it does make sense to use this library without `opm.io`? If the answer is no... ;)

>to let it fail with whatever exception

Would it make sense to add an `if not HAVE_OPM: Raise ImportError(opm is required for this function)` to any function that requires opm? (or make it slightly more elegant with a decorator). Just for the user to get some more sensible feedback? `no module named opm` would be a best case perhaps, could get worse error messages..

berland:
> But it does make sense to use this library without `opm.io`? If the answer is no... ;)

Yes, the libecl dependent parts work fine without opm.

> > to let it fail with whatever exception
> Would it make sense to add an `if not HAVE_OPM: Raise ImportError(opm is required for this function)` to any function that requires opm? (or make it slightly more elegant with a decorator). Just for the user to get some more sensible feedback? `no module named opm` would be a best case perhaps, could get worse error messages..

It would make sense, but I think it requires some refactoring to the import setup (from `__init__.py`) to make it work.

@miroine  do you want to contribute to such a thing?

The user that gets this far should be assumed to have read the paragraph in `contribution.rst` saying that they are on their own when running without opm.

lars-petter-hauge:
> The user that gets this far should be assumed to have read the paragraph in `contribution.rst` saying that they are on their own when running without opm.

I agree, as it is required from the user to specifically install without dependencies, one should really be aware that it is a custom installation.

lars-petter-hauge (approved):

lars-petter-hauge:
Should've approved this by the last comment.

PR #270, opened 2021-02-16
Solve logging bug for summary

berland:
Prior to this fix, calling `--verbose` did not print the message with which filename was written to. The index option was set correct by pure luck.

### DIFF
--- a/ecl2df/summary.py
+++ b/ecl2df/summary.py
@@ -568,7 +568,7 @@ def summary_main(args):
     )
     if sum_df.empty:
         logger.warning("Empty summary data being written to disk!")
-    write_dframe_stdout_file(sum_df, args.output, logger)
+    write_dframe_stdout_file(sum_df, args.output, index=True, caller_logger=logger)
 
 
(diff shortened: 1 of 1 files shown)


asnyv:
This is not just a logging change? 
From `index=False` (default in `write_dframe_stdout_file`) to` `index=True` is a significant difference?

berland:
> This is not just a logging change?
> From `index=False` (default in `write_dframe_stdout_file`) to` `index=True` is a significant difference?

The `logger` object was taken as the `index` argument, and was evaluated to `True`. That was the luck part (but testing guaranteed the luck).

asnyv:
> > This is not just a logging change?
> > From `index=False` (default in `write_dframe_stdout_file`) to` `index=True` is a significant difference?
> 
> The `logger` object was taken as the `index` argument, and was evaluated to `True`. That was the luck part (but testing guaranteed the luck).

hahah, ok 😂🙈

asnyv (approved):
LGTM 🎉

PR #271, opened 2021-02-18
Change column name from WATER_TOTAL to WIIP_TOTAL.

berland:
For consistency with other column names produced by fipreports, and
with the usage of fipreports in subscript's prtvol2csv.


asnyv:
LGTM 👍

rnyb (approved):

PR #274, opened 2021-02-25
Fix csv2ecl ert hook

berland:
Two commits solving two bugs:

1: csv2ecl summary did not work as an ERT forward model due to the argparse argument not being aligned. This had to be changed. Few users assumed, so this can be changed with no deprecation period. csv2ecl summary is still new.

2. cvs2ecl summary could not write to subdirectories, which is the natural setting for an  ERT forward model. Added as a feature.

### DIFF
--- a/ecl2df/summary.py
+++ b/ecl2df/summary.py
@@ -1,4 +1,5 @@
 """Provide a two-way Pandas DataFrame interface to Eclipse summary data (UNSMRY)"""
+import os
 import logging
 from pathlib import Path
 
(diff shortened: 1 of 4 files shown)


wouterjdb (approved):
LGTM 🎉

ecl2df/summary.py
```diff
@@ -694,6 +695,14 @@ def summary_reverse_main(args):
...
-    eclsum = df2eclsum(summary_df, args.output)
+    outputdir = Path(args.output).parent
+    eclbase = Path(args.output).name
+
+    # EclSum.fwrite can only write to current directory:
+    cwd = os.getcwd()
+    os.chdir(outputdir)
+    eclsum = df2eclsum(summary_df, eclbase)
     EclSum.fwrite(eclsum)
+    os.chdir(cwd)
```
lars-petter-hauge: I suggest to use a `contextmanager` to handle returning to the original directory - although no `contextmanager` will stand a chance against a `util_abort`.

I understand this is a limitation in ecl, but it should rather be fixed there (I had a short glance, and it doesn't seem to be trivial to fix as ecl sets the casename on the c-object when wanting to write, instead of passing a path)

berland: I thought of a context manager, but failed to see the upside in this situation?

PR #275, opened 2021-02-26
Ensure cwd is not manipulated in summary

berland:
EclFile.write() easily exits hard in C-code, but this fix should conserve cwd if it ever fails more softly.


lars-petter-hauge (approved):
As long as libecl does not allow to write anywhere outside of current working directory, I suppose this is the best we can do.

PR #276, opened 2021-03-01
Use pathlib for hook_implementations

berland:


ecl2df/hook_implementations/jobs.py
```diff
@@ -1,20 +1,20 @@
...
     all_files = [
-        os.path.join(resource_directory, f)
-        for f in os.listdir(resource_directory)
-        if os.path.isfile(os.path.join(resource_directory, f))
+        resource_directory / filename
+        for filename in resource_directory.glob("*")
+        if (resource_directory / filename).exists()
     ]
-    return {os.path.basename(path): path for path in all_files}
+    return {str(Path(path).name): str(path) for path in all_files}
```
lars-petter-hauge, line 17: The result of a `/` operator on a `Path` and a `string` will result in a new `Path` instance, hence no need to create another one here.

lars-petter-hauge:
Only a single comment really, otherwise looks good to me!

lars-petter-hauge (approved):

PR #277, opened 2021-03-18
Solve probable bug with include_restart

berland:
Found by code inspection. For combinations of time_index and
include_restart, the include_restart setting would probably
get incorrect due to caching

### DIFF
--- a/ecl2df/summary.py
+++ b/ecl2df/summary.py
@@ -262,11 +262,19 @@ def df(
         column_keys = [column_keys]
     if isinstance(time_index, str) and time_index == "raw":
         time_index_arg = resample_smry_dates(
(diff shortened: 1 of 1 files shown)


lars-petter-hauge:
So the `eclfiles.get_eclsum(include_restart=include_restart)` is called in three places here, two times to collect only the dates, the last one to get the data. But the first two will still create an instance of eclsum. Furthermore the call when collecting data has a try except, but this is not guarded for when retrieving dates.

Would it be possible to do the `eclfiles.get_eclsum(include_restart=include_restart)` call in the beginning? If this throws, there's no point moving forward. You'll have the object already then to get the `dates` from.

berland:
Thanks for feedback, this should be fixed now.

asnyv (approved):
LGTM

PR #284, opened 2021-04-07
Drop not-a-number data from restart/grid dataframes

berland:


vkip (approved):
Trusting you on the 'trial-and-error' :)

PR #288, opened 2021-04-18
Read zonemap functionality moved to common module

lindjoha:
I have moved the parsing of lyr files to a general function in the common module. This makes it easiser to use from outside ecl2df. Functionality in the EclFiles class is exactly as before.

### DIFF
--- a/ecl2df/common.py
+++ b/ecl2df/common.py
@@ -624,7 +624,7 @@ def stack_on_colnames(dframe, sep="@", stackcolname="DATE", inplace=True):
     return dframe
 
 
-def read_zonemap(filename):
+def parse_zonemapfile(filename: str):
     """Return a dictionary from (int) K layers in the simgrid to strings
 
     Typical usage is to map from grid layer to zone names.
@@ -645,9 +645,7 @@ def read_zonemap(filename):
         dict, integer keys which are the K layers. Every layer mentioned
             in the interval in the input file is present. Can be empty.
     """
-    assert isinstance(filename, str)
-    with open(filename, "r") as fin:
-        zonelines = fin.readlines()
+    zonelines = Path(filename).read_text().splitlines()
     zonelines = [line.strip() for line in zonelines]
     zonelines = [line.split("--")[0] for line in zonelines]
     zonelines = [line for line in zonelines if not line.startswith("#")]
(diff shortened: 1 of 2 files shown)


ecl2df/common.py
```diff
@@ -621,3 +622,47 @@ def stack_on_colnames(dframe, sep="@", stackcolname="DATE", inplace=True):
...
+
+        Args:
+            filename (str): Absolute path to a zone map file (lyr format)
+
+        Returns:
+            dict, integer keys which are the K layers. Every layer mentioned
+                in the interval in the input file is present. Can be empty.
+    """
+    assert isinstance(filename, str)
+    with open(filename, "r") as fin:
```
berland, line 648: After learning `pathlib` I prefer `Path(filename).read_text().splitlines()`

ecl2df/common.py
```diff
@@ -621,3 +622,47 @@ def stack_on_colnames(dframe, sep="@", stackcolname="DATE", inplace=True):
     del dframe["level_0"]
     dframe.index.name = ""
     return dframe
+
+def read_zonemap(filename):
```
berland, line 626: Should we call it `parse_zonemap`,  or even `parse_zonemapfile`?

ecl2df/common.py
berland: (that would also make it possible to supply `Path` objects to the function)

berland: That might also solve the error for the tests to pass.

berland:
Thanks for the contribution!

berland:
There is a tip in https://equinor.github.io/ecl2df/contribution.html#continuous-integration for how to use `pre-commit` to avoid commiting files that are not passing black and flake.

ecl2df/common.py
lindjoha: Ok, I agree!

ecl2df/common.py
lindjoha: Are you thinking that I should make a Path object an input parameter to the function? F.ex optional or instead of the string?

berland:
Can you refactor the relevant test-code in test_zonemap.py also. Now this test file should test the parsing individually from the functionality of eclfiles (which only looks up a default file).

lindjoha:
Ok, let me have a look at it.

berland:
Superseded by #289

PR #293, opened 2021-04-28
Add WELOPEN tests, improve assert

### LINKED ISSUE #291: [compdat] WELOPEN connection value of zero not supported
```
WELOPEN
    'OP'      'SHUT'    0    0    0  2* /
/

```
is seen in an Eclipse deck, but will trigger crash in ecl2df.compdat due to the zero value. The eclipse manual defines what should happen on a zero value, so this is a bug.

wouterjdb:
Closes #291; bug not reproduced - seems to work as expected. Added test for the exact case.

- Add tests for WELOPEN
- Fixes sorting issues in assert

### DIFF
--- a/tests/test_welopen.py
+++ b/tests/test_welopen.py
@@ -8,6 +8,113 @@ from ecl2df import compdat
 from ecl2df import EclFiles
 
 WELOPEN_CASES = [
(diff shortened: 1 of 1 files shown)


berland (approved):
lgtm

PR #294, opened 2021-04-29
API change: ACCURACY in equil renamed to OIP_INIT

berland:
This is in line with OPM. ACCURACY was a bad choice and custom
for ecl2df


tklov11 (approved):
LGTM

PR #296, opened 2021-04-30
Solve bug for incomplete zonemap

berland:
* Add support for specifying zonemap directly in grid.df()
* Improve testing for merge_zones

### DIFF
--- a/ecl2df/common.py
+++ b/ecl2df/common.py
@@ -295,7 +295,8 @@ def parse_opmio_tstep_rec(record):
 
 def merge_zones(df, zonedict, zoneheader="ZONE", kname="K1"):
     """Merge in a column with zone names, from a dictionary mapping
-    k-index to zone name
+    k-index to zone name. If the zonemap is not covering all
+    zones, cells will be filled with NaN.
 
     Args:
         df (pd.DataFrame): Dataframe where we should augment a column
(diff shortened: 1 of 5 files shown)


ecl2df/common.py
```diff
@@ -324,7 +325,7 @@ def merge_zones(df, zonedict, zoneheader="ZONE", kname="K1"):
     zone_df = pd.DataFrame.from_dict(zonedict, orient="index", columns=[zoneheader])
     zone_df.index.name = "K"
     zone_df.reset_index(inplace=True)
-    return pd.merge(df, zone_df, left_on=kname, right_on="K")
+    return pd.merge(df, zone_df, how="left", left_on=kname, right_on="K")
```
lindjoha, line 328: Could you have done this instead of first turning the zonedict into a dictionary? Or is it not so safe wrt missing values?
```
df[zoneheader] = df[kname].map(zonedict)
return df
```

lindjoha (commented):
I have tried to get into the code and it looks good to me. I have just left a couple of very small coments.

ecl2df/common.py
berland: Thanks, this improves the code and passes all existing and new tests!

lindjoha (approved):
I have no further comments!

PR #298, opened 2021-04-30
Error hard on lumped connections

berland:
* Support zero and negative values properly
* Add user warning in doc on limitiation in dataframe concept for well open state
* Add more context in raised exceptions by outputting all info from dataframe row.

### DIFF
--- a/docs/usage/compdat.rst
+++ b/docs/usage/compdat.rst
@@ -23,7 +23,7 @@ If you need access to WELSEGS, COMPSEGS, WSEGSICD, WSEGAICD or WSEGVALV, you
 must use the ``deck2dfs()`` function which will return a dict with dataframes
 for each of COMPDAT, and the segmentation keywords.
 
-.. warning:: When WELOPEN is in use, the dataframe can differ from Eclipsex
+.. warning:: When WELOPEN is in use, the dataframe can differ from Eclipse
   behaviour in certain circumstances. The dataframe representation from
   ``compdat`` does not separate from a "shut" well and the open-ness of its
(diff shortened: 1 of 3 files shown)


berland:
Manual states:

> If items 3 - 7 are all defaulted, the Open/Shut/Stop command applies to the well, leaving the connections
> unchanged. An attempt to open a well with no connections open results in an error condition.
> 
> If any of items 3 - 7 are set zero or positive, the command applies to the eligible connections of the well,
> leaving the status of the well itself unchanged. The eligible connections are those with I,J,K locations
> matching the specified location indices in items 3 - 5 and a completion number in the range specified by
> items 6 - 7. A zero or negative value for a location index (I, J or K) matches any value of that location
> index. Similarly a zero or negative value for a completion range lower/upper end point indicates that there
> is no lower/upper bound on the completion numbers.
> 
> To open a well and one or more of its connections requires two records, one to operate on the well and the
> other to operate on the connections.

docs/usage/compdat.rst
```diff
@@ -23,6 +23,14 @@ If you need access to WELSEGS, COMPSEGS, WSEGSICD, WSEGAICD or WSEGVALV, you
 must use the ``deck2dfs()`` function which will return a dict with dataframes
 for each of COMPDAT, and the segmentation keywords.
 
+.. warning:: When WELOPEN is in use, the dataframe can differ from Eclipsex
```
wouterjdb, line 26:
```suggestion
.. warning:: When WELOPEN is in use, the dataframe can differ from Eclipse
```

wouterjdb (approved):
Tiny typo, otherwise LGTM 👍

PR #299, opened 2021-04-30
Add test for rounding when Ecl output is generated

### LINKED ISSUE #279: Too many decimals crashes Eclipse
If numbers are given to Eclipse with too many decimals (typical for floating point Python), Eclipse will crash.

The number `1000.00000000000000000003` somewhere where a float is expected, will crash Eclipse, this is dangerous for all ecl2df code that exports Eclipse include files (not yet seen in practice for ecl2df, but in other contexts)

Eclipse probably rounds everything to 32 bit floats, so some rounding should be safe to do.

berland:
Resolves #279 , or really, "proves" that this is probably not an issue for ecl2df.


tests/test_equil.py
```diff
@@ -357,6 +358,40 @@ def test_ntequl():
...
+    [
+        (1000.0000000000003, " 1000.0 "),  # All these decimals would crash Eclipse
+        (1000.0000003, " 1000.0 "),
+        (1000.000003, " 1000.000003 "),  # Eclipse should accept this.
+        (np.float32(1000.00003), " 1000.0 "),
+        (np.float32(1000.0003), " 1000.0003"),  # can give 1000.000305
+    ],
+)
+def test_eclipse_rounding(somefloat, expected):
+    """Values in include files with a lot of decimals, like you sometimes get
```
anders-kiaer, line 372: "a lot of decimals" --> replace with exact number of decimals/digits/resolution if known?

anders-kiaer (approved):

tests/test_equil.py
berland: (I don't know it.. will need testing in Eclipse)

PR #307, opened 2021-05-11
Improve lyr parsing

lindjoha:
Added functionality to parse lyr files with color code.

Functionality is split into two functions in the common module: parse_lyrfile (returns a dictionary representation of the lyr file) and convert_lyrparse_to_zonemap. The zonemap name still refers to the layer->zone mapping as before and the get_zonemap function in eclfiles still returns the same zonemap.

Some changes has also been done in test_zonemap.py and one new test has been added.


ecl2df/common.py
```diff
@@ -627,43 +628,87 @@ def stack_on_colnames(dframe, sep="@", stackcolname="DATE", inplace=True):
...
+            output[zone_name] = zone_attributes
         except ValueError:
-            logger.error("Could not parse zonemapfile %s", filename)
+            logger.error("Could not parse lyr file %s", filename)
             logger.error("Failed on content: %s", line)
             return None
+    return output
+
+
+def convert_lyrparse_to_zonemap(lyrparse: dict) -> Dict[int, str]:
```
berland, line 696: `lyrparse` -> `lyrdict` ?

ecl2df/common.py
```diff
@@ -627,43 +628,87 @@ def stack_on_colnames(dframe, sep="@", stackcolname="DATE", inplace=True):
...
 
-    Typical usage is to map from grid layer to zone names.
 
-    The layer filename must currently follow format:
+def parse_lyrfile(filename: str) -> Dict[str, Dict[str, Any]]:
+    """Return a dictionary presentation of the lyr file.
 
-      'ZoneA' 1-4
-      'ZoneB' 5-10
+    The lyr file contains data of the following format,
```
berland, line 641: Is there an URL (resinsight.org?) that defines this format that we can refer to?

tests/test_zonemap.py
```diff
@@ -45,43 +46,44 @@ def test_nonexistingzones():
...
+    assert "Could not parse lyr file" in caplog.text
     assert "Failed on content: foo" in caplog.text
 
-    zonefile = tmpdir / "zonemap.lyr"
-    zonefile.write_text(
+    lyrfile = tmpdir / "formations.lyr"
+    lyrfile.write_text(
         """
-valid 1-2
+valid 1-2 #FFE5F7
```
berland, line 64: Add a line with a wrong colour spec also, that will fail (#FFGGHH)

ecl2df/common.py
```diff
@@ -627,43 +628,87 @@ def stack_on_colnames(dframe, sep="@", stackcolname="DATE", inplace=True):
     return dframe
 
 
-def parse_zonemapfile(filename: str):
-    """Return a dictionary from (int) K layers in the simgrid to strings
+def _is_color(input_string: str) -> bool:
+    """Check if the input string is a 6 digit hexadecimal color code"""
+    if input_string.startswith("#") and len(input_string) == 7:
```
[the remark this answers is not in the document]
lindjoha: Ok, thanks. I didn't know you could turn hexadecimal strings into int's like that

ecl2df/common.py
lindjoha: Yes, I agree. I will change it to that everywhere

ecl2df/common.py
lindjoha: Yes: https://resinsight.org/3d-main-window/formations/#formation-names-description-files-_lyr_

lindjoha: There are still some functionality we are not handling. F.ex this. That is a bit tricky to solve.
image

lindjoha: and the SVG color keyword names.

ecl2df/common.py
```diff
@@ -627,43 +628,87 @@ def stack_on_colnames(dframe, sep="@", stackcolname="DATE", inplace=True):
...
             return None
+    return output
+
+
+def convert_lyrparse_to_zonemap(lyrparse: dict) -> Dict[int, str]:
+    """Returns a layer to zone map as a dictionary
+
+    Args:
+        lyrparse (dict): dictionary representation of a lyr file coming from the
+        parse_lyrfile function
```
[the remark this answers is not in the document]
berland: I _think_ line 701 needs 4 more spaces for the sphinx-formatting to be correct.

berland:
The resinsight doc uses "layer_zone_table" for the same thing as "zonemap" here. Should we change to that terminology?

ecl2df/common.py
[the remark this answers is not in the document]
berland: or drop the check, see comment below.

berland:
ok. might be a good idea to keep as is.

ecl2df/common.py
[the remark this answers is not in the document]
lindjoha: ok. yes I see. That's why it's failing in the checks also.

ecl2df/common.py
lindjoha: It's not a single layer format, it's 3 additional layers relative to the previous line. So that adds quite a lot of logic to the code. Should we also check that zones don't have overlapping layers?

lindjoha: But we can't accept everything as an color. F.ex:
`ZoneA 3     -10`

This shouldn't be interpreted as 3 layers with colorcode "-10"

berland: good point

berland:
See also https://github.com/OPM/ResInsight/blob/5f44c994bfb51f4f861d1e5c338fcd116b5e8761/ApplicationLibCode/FileInterface/RifColorLegendData.cpp#L51  for the C++ version of the similar parser.

berland:
For validating colours (if we want to), this package should probably be used: https://pypi.org/project/colour/

ecl2df/common.py
lindjoha: I agree with your comment about making the output a list of dicts

lindjoha:
```
'Eiriksson'  1-10
 Raude    20-30
```

Resinsight does not accept formation names without quotes. Should we follow the same rule?

berland:
> Resinsight does not accept formation names without quotes. Should we follow the same rule?

I think it is okay to let `shlex` do that parsing as before (which implies that it is more flexible than ResInsight).

ecl2df/common.py
```diff
@@ -643,26 +643,28 @@ def parse_lyrfile(filename: str) -> List[Dict[str, Any]]:
     #formation-names-description-files-_lyr_
 
     The output has the following format:
-
+    ```json
```
berland, line 646: "json" here is not valid  in sphinx it seems.

ecl2df/common.py
```diff
@@ -643,26 +643,28 @@ def parse_lyrfile(filename: str) -> List[Dict[str, Any]]:
     #formation-names-description-files-_lyr_
 
     The output has the following format:
```
berland, line 645: You might only need a double colon at the end here for sphinx to take the following section as verbatim, and then 
you can have the indents as you wish.

Verify the generated doc with `python setup.py build_sphinx` and then `firefox  build/sphinx/html/index.html` from the project root.

berland:
I  wonder  if I am at back at Jens's suggestion of just two regexes and a fixed list (possibly in an extra text file?) for the colour recognition. It is only 140 lines of colours and it might be safe to assume it is static for the foreseeable future.

ecl2df/common.py
lindjoha: Ok, I will remove it. What about the quotes?

lindjoha:
Yes, I was also thinking about that. What if I make a colors module with the list and the is_color() function ?

ecl2df/common.py
lindjoha: image

lindjoha:
I'll just put it in a text file :)

ecl2df/common.py
lindjoha: I was not able to run `python setup.py build_sphinx`. See error message above

berland:
Consider case insensitive colour search

tests/test_zonemap.py
```diff
@@ -45,74 +46,123 @@ def test_nonexistingzones():
...
+    lyrfile = tmpdir / "formations.lyr"
+    lyrfile.write_text(
         """
--- Layer table for 83 layer simulation model xxxx
-'UT3_3'          1 -     15
-'UT3_2'          16 -   20
-'UT3_1'         21 -    21
-'UT2'           22 -   23
-'UT1_2'         24 -   37
-'UT1_1'         38 -   46 -- thistextisignored
```
berland, line 92: looks like this corner case has been lost in the edited version? (comment-at-end-of-line)

tests/test_zonemap.py
```diff
@@ -45,74 +46,123 @@ def test_nonexistingzones():
...
+'ZoneC'          11-15    blue
+'ZoneD'         3          #fbb
+'ZoneE'         19     -20
+'ZoneF'         21-22  CORNFLOWERBLUE
+""",
+        encoding="utf-8",
+    )
+    lyrlist = ecl2df.common.parse_lyrfile(lyrfile)
+    # 0 - ZoneA
+    assert lyrlist[0]["from_layer"] == 1
```
berland, line 111: Can all these asserts be condensed into a more explicit test, like 
```
assert lyrlist == [{"from_layer": 1, "name": ZoneA, "color": "#FFE5F7}, {....}]
```

tests/test_zonemap.py
berland: Looks like it is covered further up, in ZoneB

tests/test_zonemap.py
```diff
@@ -45,74 +46,123 @@ def test_nonexistingzones():
...
+    lyrfile = tmpdir / "formations.lyr"
+    lyrfile.write_text(
+        """
+valid 1-2 #FFE5F7
+foo   3- 4 #FFGGHH
+""",
+        encoding="utf-8",
+    )
+    assert ecl2df.EclFiles(DATAFILE).get_zonemap(str(lyrfile)) is None
+    assert "Failed on content: foo   3- 4 #FFGGHH" in caplog.text
```
berland, line 91: Does this cover the case for a misspelled color, like 'bluez'?

tests/test_zonemap.py
lindjoha: I thought it was coverd by the ZoneB line here? and the test `assert "color" not in lyrlist[1]`

```
def test_lyrlist_format(tmpdir):
    """Ensure the lyr file is parsed correctly"""
    lyrfile = tmpdir / "formations.lyr"
    lyrfile.write_text(
        """
-- Some text
'ZoneA'          1 -     5  #FFE5F7
'ZoneB'          6-     10  --no color
'ZoneC'          11-15    blue
'ZoneD'         3          #fbb
'ZoneE'         19     -20
'ZoneF'         21-22  CORNFLOWERBLUE
""",
        encoding="utf-8",
    )
    lyrlist = ecl2df.common.parse_lyrfile(lyrfile)
    # 0 - ZoneA
    assert lyrlist[0]["from_layer"] == 1
    assert lyrlist[0]["color"] == "#FFE5F7"
    assert lyrlist[0]["name"] == "ZoneA"
    # 1 - ZoneB
    assert lyrlist[1]["to_layer"] == 10
    assert "color" not in lyrlist[1]
```

tests/test_zonemap.py
lindjoha: Yes, I was wondering if that was possible. I will do it.

tests/test_zonemap.py
lindjoha: I guess not. I will add it

ecl2df/common.py
```diff
@@ -651,43 +656,107 @@ def stack_on_colnames(
     return dframe
 
 
-def parse_zonemapfile(filename: str) -> Optional[dict]:
-    """Return a dictionary from (int) K layers in the simgrid to strings
+def is_color(input_string: str) -> bool:
+    """Checks if the input string is avalid color.
```
berland, line 660: avalid -> a valid

ecl2df/common.py
```diff
@@ -651,43 +656,107 @@ def stack_on_colnames(
...
+    That is six-digit hexadecimal, three-digit hexadecimal or
+    given as a SVG color keyword name
```
berland, line 662: "as an SVG"

ecl2df/common.py
```diff
@@ -651,43 +656,107 @@ def stack_on_colnames(
...
+    """
+    if input_string.lower() in SVG_COLOR_NAMES:
```
berland, line 664: (would fail if someone puts in upper cases in the text file, but noone is supposed to do that..)

ecl2df/common.py
```diff
@@ -651,43 +656,107 @@ def stack_on_colnames(
...
+            {
+                "name": "ZoneB",
+                "span": 5,
+                "color": "red"
+            }
+        ]
 
     Args:
-        filename (str): Absolute path to a zone map file (lyr format)
+        filename (str): Absolute path to a lyr file
```
berland, line 700: `filename (str): Absolute` -> `filename: Absolute` 

sphinx autotyping takes care of stating that the type is a striing.

ecl2df/common.py
```diff
@@ -651,43 +656,107 @@ def stack_on_colnames(
...
 
     Returns:
-        dict, integer keys which are the K layers. Every layer mentioned
-            in the interval in the input file is present. Can be empty.
-    """
+        List[Dict[str, Any]]
```
berland, line 703: The type of the returned object is already inserted by sphinx autotyping. Better describe in words what is returned.

ecl2df/common.py
lindjoha: Do you think I should do: SVG_COLOR_NAMES.lower() also?

berland: If we state that the code is case-insensitive, that is maybe smart.

lindjoha: I have added it where the list is loaded in the beginning of the file

PR #308, opened 2021-05-11
Add typehinting to ecl2df

berland:


ecl2df/common.py
```diff
@@ -73,25 +78,29 @@
...
+logger: logging.Logger = logging.getLogger(__name__)
 
 
 def write_dframe_stdout_file(
-    dframe, output, index=False, caller_logger=None, logstr=None
-):
+    dframe: pd.DataFrame,
+    output: str,
+    index: bool = False,
+    caller_logger: logging.Logger = None,
```
anders-kiaer, line 90: `Optional`?

ecl2df/common.py
```diff
@@ -212,25 +226,27 @@ def ecl_keyworddata_to_df(
 
 
 def parse_opmio_deckrecord(
-    record, keyword, itemlistname="items", recordindex=None, renamer=None
-):
+    record: "opm.libopmcommon_python.DeckRecord",
+    keyword: str,
+    itemlistname: str = "items",
+    recordindex: int = None,
```
anders-kiaer, line 232: `Optional`

anders-kiaer (approved):

ecl2df/common.py
berland: Yes. Not sure why `mypy` accepts this.

anders-kiaer: I was wondering about the same - looks like an extra commandline value is needed:
https://mypy.readthedocs.io/en/stable/command_line.html#none-and-optional-handling

Maybe we should add that flag in `webviz-config`. Currently we are only using these flags over in that repository:
```
mypy --package webviz_config --ignore-missing-imports --disallow-untyped-defs --show-error-codes
```

anders-kiaer: Alternatively maybe simply use the `mypy` default practice (infer `Optional` from `None` default). I did not know it was a "feature" in `mypy`.

PR #309, opened 2021-05-18
Make `ert` an optional dependency

### LINKED ISSUE #209: Make ert an optional dependency?
...if so, this is useful for those not using ERT, but only ecl2df in :cloud: or in other contexts. ERT again depends on some big dependencies like PyQT.

anders-kiaer:
Closes #209.


.github/workflows/ecl2df.yml
```diff
@@ -24,6 +24,7 @@ jobs:
     strategy:
       matrix:
         python-version: ['3.6', '3.7', '3.8', '3.9']
+        install-ert: [true, false]
```
berland, line 27: Do we need to test this for every python-version? (thinking of electricity consumption..)

anders-kiaer: We can limit it, not sure what would be the preferred GitHub workflow syntax?

I see https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-including-additional-values-into-combinations is a concept - not entirely clear to me what happens when doing `if matrix.some_variable` in branches where the new variable is not added :thinking: 

Alternatively wright the combinations fully out.

setup.py
```diff
@@ -17,8 +17,6 @@
 SETUP_REQUIREMENTS = ["setuptools>=28", "setuptools_scm"]
 REQUIREMENTS = [
     "ecl",
-    "equinor-libres",
```
anders-kiaer, line 20: Not sure if it is correct to remove this. I did a quick search for `libres` / `res` but could not find any direct imports. Is it used directly?

berland: It probably is an indirect dependency, originally added because both ert and libres was explicitly compiled from source prior to pypi.

berland (approved):
An environmentally friendly PR! ;) 👍🏻

PR #310, opened 2021-05-20
Integrate sphinx-apidoc command into conf.py

berland:
Left margin menu in generated docs are slightly altered,
the API docs are now at the end, and stands more clearly
out.


lindjoha (approved):
Looks good to me!

PR #312, opened 2021-05-26
Add support for WLIST

### LINKED ISSUE #218: [compdat] WELOPEN does not support WLIST
If well lists are in use in WELOPEN, correct behaviour depends on the keyword WLIST to be supported.

Should WELOPEN fail hard when well lists are in use, or give warning (failing hard is maybe the correct thing to do, as the well states will be wrong)

berland:
Resolves #218

### DIFF
--- a/ecl2df/common.py
+++ b/ecl2df/common.py
@@ -68,6 +68,7 @@ for keyw in [
     "WELOPEN",
     "WELSEGS",
     "WELSPECS",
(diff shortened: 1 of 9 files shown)


lindjoha (approved):
I have tested the new functionality on a real model and it works very well.

berland:
🚢

PR #313, opened 2021-05-26
Fix unclosed resource/file warning

anders-kiaer:
Removes
```
 [...]/python3.6/site-packages/ecl2df/common.py:87: ResourceWarning:

unclosed file <_io.TextIOWrapper name='[...]/python3.6/site-packages/ecl2df/svg_color_keyword_names.txt' mode='r' encoding='UTF-8'>
```


berland (approved):
Thanks!

PR #318, opened 2021-06-10
Add note in doc on math.fsum()

### LINKED ISSUE #316: [grid] Return floating point columns as float64 for stable summation
The floating point colums returned by `grid.df()` are of type `float32`, as that is the precision used in libecl.

When arrays of float32 are summed naively, using  `.sum()`, this results in inaccuracies, see for example https://en.wikipedia.org/wiki/Kahan_summation_algorithm

If the user always uses `math.fsum()`, this is not a problem, and the user should maybe always do that.

Still, by sacrificing memory, and converting the arrays to float64, `.sum()` will be as accurate as `math.fsum()` in test cases.

As knowlegde of numerical summation problems and floating point is not widespread, this might be worth it.

berland:
Resolves #316

### DIFF
--- a/docs/usage/grid.rst
+++ b/docs/usage/grid.rst
@@ -44,8 +44,13 @@ Example computations on a grid dataframe
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 Some grid statistic operations are very neatly expressed using Python and
-Pandas. Some examples (provided that the ``drame`` object is initialized as in
+Pandas. Some examples (provided that the ``dframe`` object is initialized as in
 the topmost example):
+
+Summation of floating point numbers is difficult for computers, as summing each
+number in a sequence can lead to accumulation of roundoff errors. This is
+proven to have an impact on volumetrics computations on a grid. For this, the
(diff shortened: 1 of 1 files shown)


berland:
Needs #317 for CI to pass.

asnyv (approved):
Out of curiosity: I assume math.fsum() is slower due to a more complex summation algorithm? 
If so, do you know how much? 
At least for interactive use, you might be willing to loose some precision to gain speed 🏎 (unless your experience is that regular sum() gets terribly wrong?)

berland:
> Out of curiosity: I assume math.fsum() is slower due to a more complex summation algorithm?
> If so, do you know how much?
> At least for interactive use, you might be willing to loose some precision to gain speed 🏎 (unless your experience is that regular sum() gets terribly wrong?)

I don't know how much slower. The most accurate summation algorithm would be to sort all number ascending before summing, but that is too costly, so `fsum()` is doing some smart compromise.

When validating volumes between rms/geogrid/simgrid/prt files, numerical discrepancies easily causes a lot mental noise, probably warranting the extra cost of `fsum()`

PR #329, opened 2021-07-05
Fix non-functional stackdates

berland:
This fixes an ugly but that has been reported by an asset user, and probably warrants pushing into komodo asap.

The code that has been touched is apparently ugly, and could be a candidate for further refactoring.

### DIFF
--- a/tests/test_grid.py
+++ b/tests/test_grid.py
@@ -279,7 +279,7 @@ def test_df():
     grid_df = grid.df(eclfiles, vectors=["PRESSURE"], rstdates="last", stackdates=True)
     assert "PRESSURE" in grid_df
     assert len(grid_df.columns) == geometry_cols + 2
-    assert "DATE" in grid_df  # awaits stacking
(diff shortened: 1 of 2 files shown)


tests/test_grid.py
```diff
@@ -417,11 +420,21 @@ def test_rst2df():
     assert rst_df["DATE"].unique()[0] == "2000-01-01"
     rst_df = grid.rst2df(eclfiles, "all", stackdates=True)
     assert len(rst_df["DATE"].unique()) == len(grid.get_available_rst_dates(eclfiles))
-    assert rst_df.shape == (4 * 35817, 23 + 1)  # "DATE" is now the extra column
+
+    # "DATE" and "active" is now the extra columns:
```
asnyv, line 424:
```suggestion
    # "DATE" and "active" are now the extra columns:
```

asnyv (approved):

tests/test_grid.py
berland: fixed

PR #338, opened 2021-08-10
Close file descriptors opened by ecl

berland:


ecl2df/eclfiles.py
```diff
@@ -199,6 +210,22 @@ def get_prtfilename(self) -> str:
...
+            self._egridfile = None
+        if self._initfile is not None:
+            self._initfile = None
+        if self._eclsum is not None:
+            # This is necessary for garbage collection to close the EclSum file:
+            self._eclsum = None
+        if self._rstfile is not None:
+            self._rstfile = None
+        if self._rftfile is not None:
+            self._rftfile = None
```
anders-kiaer, line 227:
```suggestion
    def close(self) -> None:
        """Close any opened files. Most files are opened though ecl with
        an option to close the stream as possible, leaving not much work
        for this function."""
        self._egridfile = None
        self._initfile = None
        # This is necessary for garbage collection to close the EclSum file:
        self._eclsum = None
        self._rstfile = None
        self._rftfile = None
```

anders-kiaer (approved):
One possible suggestion, otherwise LGTM.

Issue #339, opened 2021-08-11
[satfunc] Support defaulted values


berland:
Single values inside a SWOF/SWFN table can be defaulted, as in:

```
SWFN
0 0 1
0.5 0.5 1*  
1 1 0
/
```
where the default should be linearly interpolated in saturation, to the value 0.5. This interpolation does not take place when a DeckItem object is asked for `get_raw_data_list()` in `ecl2df/common.py::parse_deckrecord()`. 

See also
* https://github.com/OPM/opm-common/issues/2598
* https://github.com/equinor/subscript/issues/406

PR #340, opened 2021-08-12
Support defaulted values in satfunc data

berland:
Solves #339

### DIFF
--- a/ecl2df/common.py
+++ b/ecl2df/common.py
@@ -243,7 +243,7 @@ def parse_opmio_deckrecord(
     itemlistname: str = "items",
     recordindex: Optional[int] = None,
     renamer: Optional[Union[Dict[str, str], Dict[str, Union[str, List[str]]]]] = None,
-) -> dict:
+) -> Dict[str, Any]:
     """
     Parse an opm.io.DeckRecord belonging to a certain keyword
 
@@ -263,7 +263,7 @@ def parse_opmio_deckrecord(
     if keyword not in OPMKEYWORDS:
         raise ValueError(f"Keyword {keyword} not supported by common.py")
 
-    rec_dict = {}
+    rec_dict: Dict[str, Any] = {}
 
     if recordindex is None:  # Beware, 0 is different from None here.
         itemlist = OPMKEYWORDS[keyword][itemlistname]
@@ -294,14 +294,21 @@ def parse_opmio_deckrecord(
                     rec_dict[item_name] = record[item_idx].get_data_list()
                 # (the caller is responsible for unrolling this list with
                 # correct naming of elements)
+
+                # When we parse a list, some values in it can be defaulted, for
+                # which the default values are not provided in the JSON. Return
+                # these defaulted values as NaN's for the calling code to fix.
+                for idx in range(len(record[item_idx])):
+                    if record[item_idx].__defaulted(idx):
+                        rec_dict[item_name][idx] = np.nan
         else:
             rec_dict[item_name] = jsonitem.get("default", None)
 
     if renamer:
-        renamed_dict = {}
+        renamed_dict: Dict[str, Any] = {}
         for key, value in rec_dict.items():
             if key in renamer and not isinstance(renamer[key], list):
-                renamed_dict[renamer[key]] = value
+                renamed_dict[renamer[key]] = value  # type: ignore
             else:
                 renamed_dict[key] = value
         return renamed_dict
--- a/ecl2df/common.py
+++ b/ecl2df/common.py
@@ -299,6 +299,10 @@ def parse_opmio_deckrecord(
                 # which the default values are not provided in the JSON. Return
                 # these defaulted values as NaN's for the calling code to fix.
                 for idx in range(len(record[item_idx])):
+                    # This code is using a private attribute of an
+                    # OPM DeckItem. A better solution has not yet
+                    # been found in the OPM API. See also
+                    # https://github.com/OPM/opm-common/issues/2598
                     if record[item_idx].__defaulted(idx):
                         rec_dict[item_name][idx] = np.nan
         else:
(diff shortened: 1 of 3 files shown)


ecl2df/common.py
```diff
@@ -293,14 +293,21 @@ def parse_opmio_deckrecord(
                     rec_dict[item_name] = record[item_idx].get_data_list()
                 # (the caller is responsible for unrolling this list with
                 # correct naming of elements)
+
+                # When we parse a list, some values in it can be defaulted, for
+                # which the default values are not provided in the JSON. Return
+                # these defaulted values as NaN's for the calling code to fix.
+                for idx in range(len(record[item_idx])):
```
oyvindeide, line 300: Use enumerate?

ecl2df/common.py
```diff
@@ -293,14 +293,21 @@ def parse_opmio_deckrecord(
...
+                    if record[item_idx].__defaulted(idx):
```
oyvindeide, line 301: Is double underscore intended and needed here?

tests/test_satfunc.py
```diff
@@ -310,6 +310,129 @@ def test_wrong_columns():
     assert len(wrongdf["SW"].unique()) == 3
 
 
+def test_defaulted_values():
```
oyvindeide, line 313: This should probably be split into different tests.

ecl2df/satfunc.py
```diff
@@ -120,6 +122,27 @@ def df(
     return pd.DataFrame()
 
 
+def interpolate_defaults(dframe: pd.DataFrame) -> pd.DataFrame:
```
oyvindeide, line 125: Should perhaps have a unit test?

ecl2df/common.py
berland: It is intended, I have not found another way to do it through `opm`s api.

ecl2df/common.py
berland: Would like to, but I get: `TypeError: 'opm.libopmcommon_python.DeckItem' object is not iterable`

tests/test_satfunc.py
berland: fixed.

ecl2df/satfunc.py
berland: yes. will make!

tests/test_satfunc.py
```diff
@@ -433,6 +461,158 @@ def test_defaulted_values():
...
+            pd.DataFrame(
+                [
+                    {"SW": 0, "KR": 0, "SATNUM": 1},
+                    {"SW": 0.1, "KR": 0.1, "SATNUM": 1},
+                    {"SW": 1, "KR": 1, "SATNUM": 1},
+                ]
+            ),
+        ),
+        (
+            # It will not care about column names:
```
oyvindeide, line 539: I think these are good, but personally I have started using `pytest.param` and `id=` which I think is great for readability when using `parameterize`:
```suggestion
            pytest.param(pd.DataFrame(
                [
                    {"SW": 0, "FOO": 0, "SATNUM": 1},
                    {"SW": 0.1, "SATNUM": 1},
                    {"SW": 1, "FOO": 1, "SATNUM": 1},
                ],
            pd.DataFrame(
                [
                    {"SW": 0, "FOO": 0, "SATNUM": 1},
                    {"SW": 0.1, "FOO": 0.1, "SATNUM": 1},
                    {"SW": 1, "FOO": 1, "SATNUM": 1},
                ]
       id="Will not care about column names",
        ),
```
I think the syntax is valid, but did not check it.

berland: Agree, this looks good. Will fix.

tests/test_satfunc.py
```diff
@@ -433,6 +461,158 @@ def test_defaulted_values():
...
+            None,
+            marks=pytest.mark.xfail(
+                raises=AssertionError,
+                match="Could not determine a single saturation column",
+            ),
+        ),
+        pytest.param(
+            pd.DataFrame([{"SW": 1, "SG": 1}]),
+            None,
+            marks=pytest.mark.xfail(
```
oyvindeide, line 604: I am a bit torn on this, I like the compactness. Personally I would put valid/invalid in two separate functions, but will leave it up to you.

ecl2df/common.py
oyvindeide: Is the `__defaulted` a part of the published api?

tests/test_satfunc.py
berland: Just separate out `xfail` to e.g. `test_interpolate_defaults_xfails()` or also to rewrite the expected failures to something that "passes" pytest?

oyvindeide: > also to rewrite the expected failures to something that "passes" pytest?

That is my personal preference, for example:
```
@pytest.mark.parametrize(
    "error, error_msg",
    [
        (ValueError, "something"),
        (TypeError, "something"),
    ],
)
def test_my_func(error, error_msg):
    with pytest.raises(error, match=error_msg):
        raise ValueError("something")
``` 
but again, it is my personal preference, so up to you.

berland: I think I will skip that for now, as it would imply also rewriting other existing test functions.

ecl2df/common.py
berland: Not quite sure, but have not gotten any hint to other and better alternatives in https://github.com/OPM/opm-common/issues/2598

oyvindeide: Ok, then I think that information would be good to add in a comment and/or the commit body.

berland: Added in comment.

oyvindeide: Technically not private but name mangling (though they could have intended it as private :shrug:)

oyvindeide (approved):
Dont think I have write access, but I approve!

PR #358, opened 2021-09-16
wip: Add support for pyarrow

berland:
Based on smry2arrow in webviz-subsurface.


berland:
ping @anders-kiaer

anders-kiaer:
Thanks for starting the 💭 process on this @berland - we should probably aim for investigating moving the job/code from `webviz-subsurface` to `ecl2df` in the not too distant future. The first full stack iteration is soon finished over at `webviz-subsurface`, and we can try implement the learnings into `ecl2df`. Maybe we should take a short session on this @sigurdp @berland in a couple of weeks on how best to possibly move the ERT job/code to this repository + maintain the code + cover any Webviz requirements?

ecl2df/summary.py
```diff
@@ -422,6 +424,30 @@ def df(
     return dframe
 
 
+def _df2pyarrow(dframe) -> pyarrow.Table:
+    """Construct a Pyarrow table from a dataframe, conserving metadata"""
+
+    # Datatypes to use for DATE column and all the numeric columns
+    # dt_timestamp_ms = pyarrow.timestamp("ms")
+    dt_timestamp_ms = pyarrow.timestamp("ns")  # fixme!
```
berland, line 432: yes, fixme. Probably need to detect the timestamp-type in the dataframe or always convert to milliseconds if we so decide.

ecl2df/summary.py
```diff
@@ -422,6 +424,30 @@ def df(
...
+    dt_float32 = pyarrow.float32()
+
+    # Build table column by column in order to conserve metadata.
+
+    field_list: List[pyarrow.Field] = []
+    field_list.append(pyarrow.field("DATE", dt_timestamp_ms))
+    column_arrays = [dframe.index.to_numpy()]
+
+    for colname in dframe.columns:
+        field_metadata = {b"smry_meta": json.dumps(dframe.attrs["meta"][colname])}
```
berland, line 442: the `smry` part in  `smry_meta` should not be there, as it is not generic enough, and it should be known from the context that this is a smry file.

berland: `meta` is maybe sufficient?

sigurdp: I feel that the dictionary key should be something more specific/descriptive than just `meta` to avoid confusion. 
The entire dictionary is already being stored/accessed on `pyarrow.Field`'s `metadata` property so if we just call it `meta` we wind up with: `some_field.metadata["meta"]["is_total"]` etc.

berland: Changed in the current commit  so that you get direct access to the dictionary when accessing the field metadata.

setup.py
```diff
@@ -63,6 +64,7 @@
         "console_scripts": [
             "csv2ecl=ecl2df.csv2ecl:main",
             "ecl2csv=ecl2df.ecl2csv:main",
+            "ecl2arrow=ecl2df.ecl2csv:main",
```
anders-kiaer, line 67: `ecl2arrow` == `ecl2csv` in the terminal? Does that mean that `ecl2arrow` by default gives `csv`?

berland: There is a trick here that solves that: https://github.com/equinor/ecl2df/pull/358/files#diff-6eaf200e2166a36db9538a8a79197b6a69d080a1046ea080600c8601f5182e2cR289

anders-kiaer (approved):

PR #364, opened 2021-09-27
Fixed gruptree export for the EIGHTCELLS dataset

### LINKED ISSUE #356: [gruptree] Does not yield tree with only WELSPECS
Commit 63fffaf induces a change in gruptree such that it gives an empty dataframe output from the simple Eightcells model that is made in #354 .

Ping @lindjoha

lindjoha:
Fixes #356.

The EIGHTCELLS dataset has defined WELSPECS, but not GRUPTREE. The code was ignoring the WELSPECS entry in this case, which can be considered a bug.

In addition to letting the WELSPECS entry be written to the gruptree export, a GRUPTREE entry will be written connecting the WELSPECS well-node to the FIELD group-node. Also the FIELD GRUPTREE entry (without parent) will be written to the export.

### DIFF
--- a/ecl2df/gruptree.py
+++ b/ecl2df/gruptree.py
@@ -197,10 +197,7 @@ def _write_edgerecords(
         ("GRUPTREE", "GRUPNET", "WELSPECS"),
         ("BRANPROP", "NODEPROP", "WELSPECS"),
     ]:
-        if (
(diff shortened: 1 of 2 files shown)


berland:
Squash before review?

lindjoha:
> Squash before review?

Now the commits are squashed and I have also rebased on master.

berland (approved):
LGTM

PR #365, opened 2021-09-30
Explicit testing of summary paramfile quirks

berland:
Also use mocker.patch instead of punching sys.argv

### DIFF
--- a/tests/test_summary.py
+++ b/tests/test_summary.py
@@ -236,6 +236,52 @@ def test_paramsupport(tmp_path, mocker):
     parametersyml.unlink()
 
 
+def test_paramsupport_explicitfile(tmp_path, mocker):
+    """Test explicit naming of parameters file from command line.
+
+    This is a little bit tricky because the parameter file is assumed to be
+    relative to the DATA file, not to working directory unless it is absolute."""
+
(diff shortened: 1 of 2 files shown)


tests/test_summary.py
```diff
@@ -236,6 +236,52 @@ def test_paramsupport(tmp_path, mocker):
...
+            str(randomparamfile),  # Absolute filepath
+        ],
+    )
+    ecl2csv.main()
+    assert pd.read_csv(tmpcsvfile)["FOO"].unique() == ["barrbarr"]
+    assert pd.read_csv(tmpcsvfile)["COM"].unique() == [1234]
+
+    # If we now change to tmp_path and give a relative filename to the parameter file,'
+    # it will not be found:
+    os.chdir(tmp_path)
```
anders-kiaer, line 267: Does this have a side effect of changing the cwd for all other tests as well? If so, maybe use it as a context manager or restore current cwd after this test?

anders-kiaer (approved):

tests/test_summary.py
berland: I am not 100% sure whether pytest reinstates `cwd`  for every test, but every test in `ecl2df` is written to not rely on `cwd` (the test already have to deal with `cwd` being `ecl2df`` or `ecl2df/tests`.

PR #368, opened 2021-09-30
Use tmp path

berland:


jcrivenaes (approved):
LGTM (looks like pure refactoring only,

PR #371, opened 2021-10-08
Treat WELOPEN on wells different than on connections, fixing bug

### LINKED ISSUE #222: [compdat] STOP is not a valid state for a completion
When using something like 
```
WELOPEN 
SOMEWELL STOP /
/
```
The connections/completions get the state `STOP` from `ecl2df`, whilst they in reality are `OPEN`. Crossflow can still occur, well is closed above perforations.

Similarily, if applied directly to connections/completions
```
WELOPEN 
SOMEWELL STOP 1 1 4 6 /
/
```
They also get the state `STOP`, whilst in Eclipse they would in reality be `SHUT`

`STOP` is only a state in Eclipse for wells, not completions/connections

berland:
Resolves #222 

This PR features:
* Fully ignores WELOPEN records where items 3-7 are defaulted, these statements are irrelevant to the connections
* STOP on connections are mapped to SHUT, which is equivalent for connections.


lindjoha (approved):
LGTM

PR #376, opened 2021-10-12
Remove duplicated summary columns.

berland:
This is not a problem pr. now for CSV output, but will
be a later problem when dataframes are written to arrow format.


ecl2df/summary.py
```diff
@@ -422,6 +422,17 @@ def df(
...
             dframe.index = pd.to_datetime(dframe.index)
+
+    # Remove duplicated column names. These will occur from libecl
+    # when the user has repeated vector names in the summary SECTION
+    dupes = dframe.columns.duplicated()
+    if dupes.any():
+        logger.warning(
+            "Duplicated columns detected, check your DATA file "
+            "for repeated vectors in the SUMMARY section"
+        )
```
anders-kiaer, line 433: Nice :clap:

ecl2df/summary.py
```diff
@@ -422,6 +422,17 @@ def df(
...
+        logger.warning("Duplicates: %s", str(list(dframe.columns[dupes])))
```
anders-kiaer, line 434:
```suggestion
        logger.warning("Duplicates: %s", list(dframe.columns[dupes]))
```
Or do we need to cast manually to string?

berland: Not sure.. Lets try without 😎

PR #379, opened 2021-10-18
summary with no column_key matches returns df without columns (only index)

### LINKED ISSUE #378: [summary] summary.df fails when there are no matches on column_keys
summary.df fails if there are no matches on column_keys, but it would be better if it returns an empty dataframe with only index (dates).

lindjoha:
Fixes #378

### DIFF
--- a/tests/test_summary.py
+++ b/tests/test_summary.py
@@ -86,9 +86,9 @@ def test_df_column_keys():
     assert set(sumdf.columns) == {"FOPT", "FOPR"}
     assert set(sumdf.attrs["meta"].keys()) == {"FOPT", "FOPR"}
 
-    sumdf = summary.df(EclFiles(REEK), column_keys=["BOGUS"])
-    assert sumdf.columns.empty
-    assert not sumdf.index.empty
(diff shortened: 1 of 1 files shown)


lindjoha:
> Is there a non-empty index returned, can you confirm that in tests?

I have added one more assert: `assert not sumdf.index.empty`

tests/test_summary.py
```diff
@@ -86,8 +86,9 @@ def test_df_column_keys():
     assert set(sumdf.columns) == {"FOPT", "FOPR"}
     assert set(sumdf.attrs["meta"].keys()) == {"FOPT", "FOPR"}
 
-    with pytest.raises(ValueError, match="No valid key"):
-        summary.df(EclFiles(REEK), column_keys=["BOGUS"])
+    sumdf = summary.df(EclFiles(REEK), column_keys=["BOGUS"])
+    assert sumdf.columns.empty
+    assert not sumdf.index.empty
```
berland, line 91: something like `assert all(sumdf_no_keys.index == sumdf.index)` would be even more precise

lindjoha: Deleted the previous comment. Found a better solution

PR #381, opened 2021-10-20
Refactor construction of Eclipse input tables

berland:
The code in equil.py for printing an Eclipse input table has been generalized so it can be used for 
other keywords.


berland:
Comments:
* Statements valid for py36/pandas 1.1.5 can be removed as of now
* Only the equil module is directly affected by this PR, and should be tested sufficiently already. 
* This PR adds additional tests on the table generator using COMPDAT as an example, which is relevant for future feature enhancements to the COMPDAT module (and others).

lindjoha (approved):

PR #389, opened 2021-10-26
Handling of defaulted coordinates in WELOPEN (item 3-5)

### LINKED ISSUE #334: [compdat] Improve handling of default values in WELOPEN
Default values (zero or negative) for item 3-7 in WELOPEN is not properly handled. Setting all of them to zero works, but setting some of them to zero is not implemented. F.ex:

```
COMPDAT
     'OP1' 1 1 1 1 'OPEN' /
     'OP1' 2 1 1 1 'OPEN' /
     'OP1' 1 2 1 1 'OPEN' /
    /
    WELOPEN
     'OP1' 'SHUT' 0 1 1 /
    /
```
This should SHUT the two first connections in COMPDAT, but not the last.

Default values for completions in welopen (item 6-7) is not implemented either.

lindjoha:
Fixes #334.

This PR implements the handling of default coordinates, item 3-5, in WELOPEN. The compdat dataframe is filtered to find all connections (in the well) that is matching the coordinates that are not defaulted.

### DIFF
--- a/ecl2df/compdat.py
+++ b/ecl2df/compdat.py
@@ -328,11 +328,12 @@ def expand_welopen_defaults(
             # If no coordinates are defaulted, there is nothing to expand.
             exp_welopen.append(row)
         else:
-            # If some of the coordinates is defaulted then we filter the
+            # If some of the coordinates are defaulted then we filter the
             # compdat dataframe to find the matching connections and expand
             # the WELOPEN row with those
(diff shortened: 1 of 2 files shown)


berland:
Approved, acknowledging that there are issues with how the welopen code assumes that the compdat dataframe always carry full state at every present timestep in the dataframe. Good if an issue can be written that can display what is problematic.

berland (approved):

PR #394, opened 2022-01-17
Pin opm to 2021.04

asnyv:


asnyv:
Temporary pin to pass tests and enable a release of `v0.16.0`

setup.py
```diff
@@ -18,7 +18,7 @@
 REQUIREMENTS = [
     "ecl",
     "numpy",
-    "opm>=2020.10.2",  # NB: Pypi versions.
+    "opm==2021.04",  # NB: Pypi versions.
```
berland, line 21: Looking at https://pypi.org/project/opm/#history, I would guess 2021.04.0 is the correct? Or it might find it still?

asnyv:
Hmm, yes seems like you are right. Quite messy version numbering in `opm` it seems. `2021.04` worked though, but I can of course change it 😅
image

setup.py
asnyv: Hmm, yes seems like you are right. Quite messy version numbering in `opm` it seems. `2021.04` worked though, but I can of course change it 😅
image

berland (approved):
LGTM!

PR #400, opened 2022-03-15
Adapt tests to run on a Jenkins test node

berland:
Copied setup from subscript repository

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,0 +1,11 @@
+from pathlib import Path
+
+import pytest
+
+import ecl2df
+
+
+@pytest.fixture
+def path_to_ecl2df():
+    """Path to installed ecl2df module"""
+    return Path(ecl2df.__file__).parent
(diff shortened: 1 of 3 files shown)


berland:
See https://github.com/equinor/subscript/commit/36992871aa0e419c9a42556ccbf8034e53cd1aa9

tests/conftest.py
```diff
@@ -0,0 +1,11 @@
...
+
+import pytest
+
+import ecl2df
+
+
+@pytest.fixture
+def path_to_ecl2df():
+    """Path to installed ecl2df module"""
+    return Path(ecl2df.__file__).parent
```
xjules, line 11: Should we consider absolute path?

berland: `__file__` is (almost) always absolute, since Python 3.4 it seems.

sondreso (approved):
LGTM!

PR #403, opened 2022-04-26
Complump defaults set to 0

lindjoha:
The default values of I, J, K1 and K2 in COMPLUMP should be 0 according to the manual. Defaulted values are still not implemented in the compdat module, but this give a better error message.


berland:
This change should be propagated to https://github.com/OPM/opm-common/blob/master/src/opm/input/eclipse/share/keywords/000_Eclipse100/C/COMPLUMP as well.

lindjoha:
@berland. I have made a PR to fix this also in opm-common: https://github.com/OPM/opm-common/pull/3006

lindjoha:
@berland  The change is now merged into opm-common:
https://github.com/OPM/opm-common/blob/master/src/opm/input/eclipse/share/keywords/000_Eclipse100/C/COMPLUMP

berland (approved):
Lgtm!

PR #404, opened 2022-05-23
New module `wellcompletiondata`

lindjoha:
New module `wellcompletiondata ` that uses both the `compdat` and `wellconnstatus` modules and combines this data with a layer->zone mapping `.lyr` file. The output is a table of well completion data on zone level with the following columns: WELL, ZONE, DATE, KH and OP/SH.

This is functionality was previously implemented in the `WellCompletion` plugin in `webviz-subsurface`. The reason for moving it here is to improve speed and prepare for additional functionality in the plugin. An additional advantage is that it's not necessary to export `compdat` and `wellconnstatus` to use the plugin, so less data might be exported. AND it will be easier to use the `wellconnstatus` functionality.

### DIFF
--- a/ecl2df/compdat.py
+++ b/ecl2df/compdat.py
@@ -978,7 +978,11 @@ def compdat_main(args):
     write_dframe_stdout_file(compdat_df, args.output, index=False, caller_logger=logger)
 
 
-def df(eclfiles: EclFiles, initvectors: Optional[List[str]] = None) -> pd.DataFrame:
+def df(
+    eclfiles: EclFiles,
+    initvectors: Optional[List[str]] = None,
+    zonemap_filename: Optional[str] = None,
+) -> pd.DataFrame:
     """Main function for Python API users
 
     Supports only COMPDAT information for now. Will
@@ -995,7 +999,7 @@ def df(eclfiles: EclFiles, initvectors: Optional[List[str]] = None) -> pd.DataFr
             eclfiles, compdat_df, initvectors, ijknames=["I", "J", "K1"]
         )
 
-    zonemap = eclfiles.get_zonemap()
+    zonemap = eclfiles.get_zonemap(zonemap_filename)
     if zonemap:
         logger.info("Merging zonemap into compdat")
         compdat_df = merge_zones(compdat_df, zonemap)
--- a/ecl2df/compdat.py
+++ b/ecl2df/compdat.py
@@ -986,7 +986,8 @@ def df(
     """Main function for Python API users
 
     Supports only COMPDAT information for now. Will
-    add a zone-name if a zonefile is found alongside
+    add a zone-name if a zonefile is found alongside.
+    If a zonemap is passed it will override the zonefile.
 
     Returns:
         pd.Dataframe with one row pr cell to well connection
@@ -1000,7 +1001,7 @@ def df(
         )
 
     if zonemap is None:
-        # If no zonemap is submittet, search for zonemap in default location
+        # If no zonemap is submitted, search for zonemap in default location
         zonemap = eclfiles.get_zonemap()
 
     if zonemap:
--- a/ecl2df/compdat.py
+++ b/ecl2df/compdat.py
@@ -981,7 +981,7 @@ def compdat_main(args):
 def df(
     eclfiles: EclFiles,
     initvectors: Optional[List[str]] = None,
-    zonemap_filename: Optional[str] = None,
+    zonemap: Optional[Dict[int, str]] = None,
 ) -> pd.DataFrame:
     """Main function for Python API users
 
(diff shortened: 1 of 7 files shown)


ecl2df/compdat.py
```diff
@@ -978,7 +978,11 @@ def compdat_main(args):
     write_dframe_stdout_file(compdat_df, args.output, index=False, caller_logger=logger)
 
 
-def df(eclfiles: EclFiles, initvectors: Optional[List[str]] = None) -> pd.DataFrame:
+def df(
+    eclfiles: EclFiles,
+    initvectors: Optional[List[str]] = None,
+    zonemap_filename: Optional[str] = None,
```
berland, line 984: this should probably be a `dict` (?) with the actual zonemap.

lindjoha: Yes, I can do that. But it would require bigger changes to the `compdat` code, since it is already looking for a file named `zones.lyr`.

lindjoha: Ok, no I guess it can be solved quite easily. I will add that then.

ecl2df/compdat.py
```diff
@@ -978,7 +978,11 @@ def compdat_main(args):
     write_dframe_stdout_file(compdat_df, args.output, index=False, caller_logger=logger)
 
 
-def df(eclfiles: EclFiles, initvectors: Optional[List[str]] = None) -> pd.DataFrame:
+def df(
+    eclfiles: EclFiles,
+    initvectors: Optional[List[str]] = None,
+    zonemap: Optional[Dict[int, str]] = None,
```
berland, line 984: Can you add a description of the arguments in the docstring (stating that this option for `zonemap` is for overriding the zonemap it otherwise will extract from `eclfiles`)

ecl2df/compdat.py
```diff
@@ -995,7 +999,10 @@ def df(eclfiles: EclFiles, initvectors: Optional[List[str]] = None) -> pd.DataFr
             eclfiles, compdat_df, initvectors, ijknames=["I", "J", "K1"]
         )
 
-    zonemap = eclfiles.get_zonemap()
+    if zonemap is None:
+        # If no zonemap is submittet, search for zonemap in default location
```
berland, line 1003: submittet -> submitted

ecl2df/ecl2csv.py
```diff
@@ -207,6 +207,14 @@ def get_parser() -> argparse.ArgumentParser:
         formatter_class=argparse.ArgumentDefaultsHelpFormatter,
         description=("Each row represents an edge in the GRUPTREE at a specific date."),
     )
+    subparsers_dict["wellcompletiondata"] = subparsers.add_parser(
+        "wellcompletiondata",
+        help="Extract well completion data",
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter,
+        description=(
+            "Each row represents a completion, aggregated from layer to zone."
```
berland, line 215: Testing my understanding of this description: Is this the same as "Each row represents an aggregated completion over a zones"?

ecl2df/wellcompletiondata.py
```diff
@@ -0,0 +1,248 @@
+"""Aggregates completion data from layer to zone"""
```
berland, line 1: A thought: Should the module be renamed, if it always aggregates to zones? Otherwise I would head here to get all possible data on wellcompletions (which are cell by cell), but would be disappointed if it could not give me individual well completions?

ecl2df/wellcompletiondata.py
```diff
@@ -0,0 +1,248 @@
...
+        wellconnstatus_df = wellconnstatus.df(eclfiles)
+        compdat_df = _merge_compdat_and_connstatus(compdat_df, wellconnstatus_df)
+
+    return _aggregate_layer_to_zone(compdat_df)
+
+
+def _excl_well_startswith(
+    compdat_df: pd.DataFrame, excl_well_startswith: str
+) -> pd.DataFrame:
+    "Filters out rows where the well name startswith a given string"
```
berland, line 72: `"` -> `"""`

berland: startswith -> starts with

ecl2df/ecl2csv.py
lindjoha: Yes, I guess, or maybe aggregated completion FOR zones ? For each group of equal well, date and zone, the compdat rows are aggregated to one row.

ecl2df/wellcompletiondata.py
lindjoha: Yes, names are always difficult, I have been trying to make a distinction between _connection_ (which is per cell, like in the wellconnstatus module) and _completion_ that is something more aggregated. But what could be a good name: _zonecompletion_ ?

berland: Ah, so if we just step outside the discretized world of the current grid, "a single completion" is maybe more thought of relating to one zone. Yes, keep it.

lindjoha: And it is also consistent with webviz where the plugin is called WellCompletion and it vizualises the data per zone. Good, I'll keep it then.

anders-kiaer:
Looks very good. Does this "automatically" also become available through (the) ERT job?

anders-kiaer (approved):

lindjoha:
> Looks very good. Does this "automatically" also become available through (the) ERT job?

Yes, since it it added to the ecl2csv.py file, this will happen automatically.

PR #405, opened 2022-05-30
Set Sphinx language to en

lindjoha:
Sphinx returns a warning (treated as an error) if the `language` parameter is set to  `None`. This is fixed by setting `language` to `en`.


berland (approved):
👍🏻

PR #409, opened 2022-07-06
`Wellcompletiondata` metadata

lindjoha:
Added metadata to the `wellcompletiondata` module.

### DIFF
--- a/ecl2df/wellcompletiondata.py
+++ b/ecl2df/wellcompletiondata.py
@@ -2,6 +2,7 @@
 
 import argparse
 import logging
+from enum import Enum
 from pathlib import Path
 from typing import Any, Dict, List, Optional
 
@@ -15,6 +16,13 @@ from ecl2df.eclfiles import EclFiles
 from .common import write_dframe_stdout_file
 
 logger = logging.getLogger(__name__)
+
+
+class EclipseUnitSystem(str, Enum):
+    METRIC = "METRIC"
+    FIELD = "FIELD"
+    LAB = "LAB"
+    PVTM = "PVT-M"
 
 
 def df(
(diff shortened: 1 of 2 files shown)


ecl2df/wellcompletiondata.py
```diff
@@ -63,7 +71,45 @@ def df(
...
+def _get_metadata(eclfiles: EclFiles) -> Dict[str, Dict[str, Any]]:
+    """Provide metadata for the well completion data export"""
+    meta: Dict[str, Dict[str, Any]] = {}
+    unitsystem = _get_ecl_unit_system(eclfiles)
+    kh_units = {
+        EclipseUnitSystem.METRIC: "mDm",
+        EclipseUnitSystem.FIELD: "mDft",
+        EclipseUnitSystem.LAB: "mDcm",
+        EclipseUnitSystem.PVTM: "mDm",
+    }
```
anders-kiaer, line 109: Should this also be an `Enum` at the top of the file?

ecl2df/wellcompletiondata.py
```diff
@@ -63,7 +71,45 @@ def df(
...
+    unit_systems = [unitsystem.value for unitsystem in EclipseUnitSystem]
+    for keyword in eclfiles.get_ecldeck():
+        if keyword.name in unit_systems:
+            return EclipseUnitSystem(keyword.name)
+    return EclipseUnitSystem.METRIC
+
+
+def _get_metadata(eclfiles: EclFiles) -> Dict[str, Dict[str, Any]]:
+    """Provide metadata for the well completion data export"""
+    meta: Dict[str, Dict[str, Any]] = {}
```
anders-kiaer, line 102: Could the `Any` be changed to
```suggestion
    meta: Dict[str, Dict[str, str]] = {}
```
or does it/should it be `Any`? If `kh_units` become an `Enum`, maybe it can be type hinted even stronger (https://stackoverflow.com/questions/52624736/type-annotations-for-enum-attribute).

anders-kiaer (commented):
LGTM! A couple of small questions only.

ecl2df/wellcompletiondata.py
lindjoha: Do you mean like this. I still need to keep the mapping somewhere, or do you have another idea?

```
    meta: Dict[str, Dict[str, Any]] = {}
    unitsystem = _get_ecl_unit_system(eclfiles)
    kh_units = {
        EclipseUnitSystem.METRIC: KHUnit.METRIC,
        EclipseUnitSystem.FIELD: KHUnit.FIELD,
        EclipseUnitSystem.LAB: KHUnit.LAB,
        EclipseUnitSystem.PVTM: KHUnit.PVTM,
    }
    meta["KH"] = {}
    meta["KH"]["unit"] = kh_units[unitsystem].value
    return meta
```

ecl2df/wellcompletiondata.py
```diff
@@ -17,6 +18,13 @@
 logger = logging.getLogger(__name__)
 
 
+class EclipseUnitSystem(str, Enum):
```
berland, line 21: Do you gain something from subclassing form `str` also?

tests/test_wellcompletiondata.py
```diff
@@ -89,6 +89,19 @@ def test_df2pyarrow():
     pd.testing.assert_frame_equal(df, _df2pyarrow(df).to_pandas(), check_like=True)
 
 
+def test_metadata():
+    """Test that the KH column has metadata and that unit is mDm"""
+    eclfiles = EclFiles(EIGHTCELLS)
+    df = wellcompletiondata.df(
+        eclfiles, zonemap=EIGHTCELLS_ZONEMAP, use_wellconnstatus=False
+    )
+    table = _df2pyarrow(df)
```
berland, line 98: Test the dataframe as well for metadata, not only how it gets processed to pyarrow?

ecl2df/wellcompletiondata.py
lindjoha: Yes, because I do this in line 100:

`            return EclipseUnitSystem(keyword.name)`

tests/test_wellcompletiondata.py
lindjoha: Am I not doing that in the previous test: `test_df2arrow`? Should I merge the two tests maybe?

ecl2df/wellcompletiondata.py
lindjoha: I don't think I can export an `Enum` as metadata?

I can make it `str` in this case, I guess I was thinking about other possible metadata.

tests/test_wellcompletiondata.py
lindjoha: Oh, you mean the metadata in the dataframe. Yes I see, I will do that.

ecl2df/wellcompletiondata.py
lindjoha: I have changed it to str

anders-kiaer (approved):

Issue #391, opened 2021-11-18
[summary] Warn about sub-second timesteps


berland:
It is possible in Eclipse to take sub-second timesteps that are exported to UNSMRY-files. Sub-second accuracy seems not supported in `libecl`, and in `ecl2df` it will end up as a repeated time index at the same second. This easily gets the user into trouble, and at least a warning could be issued when this (sub-second timestep) seems to be the case.

asnyv:
Actually, the issue seems to be beyond `libecl` support for sub-second accuracy. In fact it seems like Eclipse can take timesteps that are shorter than the accuracy allocated to the `TIME` vector (which at least for the `METRIC` unit system is in days) of Eclipse type "REAL" equivalent to float32. Typically a scenario of convergence issues. 

So there may actually be multiple entries with the same timestamp in the Eclipse output data. The only remaining in that case is as I see it to delete timesteps somewhere in the pipeline.

Considering the rates and pressures reported in the Eclipse output are valid backwards in time, I would assume that it is preferable to keep the first occurance and then delete the following entries if they have the same value for `TIME`, as they are only valid for an infinitesimal timestep. 

The easiest later on is of course to delete them as far up in the pipeline as possible, but there could maybe be an issue if we need consistency between number of timesteps and an index if we delete e.g. in libecl during initialization of the summary object?

asnyv:
Funny enough, if you have the `TIMESTEP` vector, the `TIMESTEP` could possibly be used to separate the different steps as the absolute values are smaller than for the `TIME` vector and thus the resolution is higher. Apparently Eclipse internals must operate on higher timestep resolution than what is exported to the output. I don't think `TIMESTEP` is a default vector though.

PR #412, opened 2022-08-11
Improve precision of datetime index using TIMESTEP

asnyv:
Enhancement to tackle the issue described in #391


sigurdp (approved):
Good to go

sigurdp (approved):

ecl2df/summary.py
```diff
@@ -435,6 +432,49 @@ def df(
         )
         logger.warning("Duplicates: %s", list(dframe.columns[dupes]))
         dframe = dframe.loc[:, ~dframe.columns.duplicated()]
+
+    index_duplicates = dframe.index.duplicated(keep="first")
+    if any(index_duplicates):
```
berland, line 437: Perhaps this should be separated out to a testable auxiliary function?

asnyv: Done, also wrote some tests (including some additional testdata), though they run on the full `df()` function

ecl2df/summary.py
```diff
@@ -435,6 +431,55 @@ def df(
...
+
+    if datetime is True:
+        if dframe.index.dtype == "object":
+            dframe.index = pd.to_datetime(dframe.index)
+
+    return dframe
+
+
+def _ensure_unique_datetime_index(dframe: pd.DataFrame) -> pd.DataFrame:
+    index_duplicates = dframe.index.duplicated(keep="first")
```
berland, line 445: docstring?

ecl2df/summary.py
```diff
@@ -435,6 +431,55 @@ def df(
...
+        if "TIMESTEP" in dframe:
+            logger.info(
+                "Dataframe of smry data contained duplicate timestamps due to limited",
+                " output resolution. Vector TIMESTEP exists, utilizing it to create "
+                "discrete timestamps",
+                f"Original duplicates were:{index_duplicate_log_string}",
+            )
+            index_as_list = dframe.index.to_list()
+
+            if dframe.attrs["meta"]["TIMESTEP"]["unit"] == "DAYS":
```
berland, line 460: I would guess you get a "KeyError" here if the unit is not available, so you will probably not go to the ValueError below. (?)

tests/test_summary.py
```diff
@@ -33,6 +33,12 @@
 TESTDIR = Path(__file__).absolute().parent
 REEK = str(TESTDIR / "data/reek/eclipse/model/2_R001_REEK-0.DATA")
 EIGHTCELLS = str(TESTDIR / "data/eightcells/EIGHTCELLS.DATA")
+SHORT_STEP = str(TESTDIR / "data/timesteps/SHORT_STEP")
```
berland, line 36: Have these been made with EIGTHCELLS? Could these example UNSMRY be made reproducible?

ecl2df/summary.py
asnyv: The idea was to catch a scenario where unit exists, but not equal to DAYS or HOURS. It was not to catch a scenario where the unit key was missing. Think the existence of the unit key is pretty much ensured by running the smry_meta command and adding it to the dataframe (currently lines 418-422)

ecl2df/summary.py
asnyv: 😴  
but ok then 😉

tests/test_summary.py
asnyv: Was made with the reek model, but doesn't really matter. Only thing was to have a schedule with forced steps, and as stripped UNSMRY files as possible. In the scenario without TIMESTEP: no vectors where requested in the SUMMARY section. In the scenarios with TIMESTEP: TIMESTEP was the only requested vector in the SUMMARY section. If I am to make it reproducible, I guess the EIGHTCELLS is better to make it light

ecl2df/summary.py
asnyv: done

berland (approved):

PR #413, opened 2022-08-16
Replace `ert_shared` with `ert.shared` following upstream change in `ert`

asnyv:
See: https://github.com/equinor/ert/pull/3752

### DIFF
--- a/ecl2df/hook_implementations/jobs.py
+++ b/ecl2df/hook_implementations/jobs.py
@@ -4,8 +4,8 @@ from pathlib import Path
 from pkg_resources import resource_filename
 
 try:
-    from ert_shared.plugins.plugin_manager import hook_implementation
-    from ert_shared.plugins.plugin_response import plugin_response
+    from ert.shared.plugins.plugin_manager import hook_implementation
+    from ert.shared.plugins.plugin_response import plugin_response
 except ModuleNotFoundError:
     # ert is not installed - use dummy/transparent function decorators.
     def hook_implementation(func):
--- a/setup.py
+++ b/setup.py
@@ -36,7 +36,11 @@ DOCS_REQUIREMENTS = [
     "sphinx-autodoc-typehints",
     "sphinx_rtd_theme",
 ]
-EXTRAS_REQUIRE = {"tests": TEST_REQUIREMENTS, "docs": DOCS_REQUIREMENTS, "ert": ["ert"]}
+EXTRAS_REQUIRE = {
+    "tests": TEST_REQUIREMENTS,
+    "docs": DOCS_REQUIREMENTS,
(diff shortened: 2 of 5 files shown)


asnyv:
should we merge prior to official release in ert @berland? (actually, the tests pass even with ert `2.37`. I haven't looked more into it though.

ecl2df/hook_implementations/jobs.py
```diff
@@ -4,8 +4,8 @@
 from pkg_resources import resource_filename
 
 try:
-    from ert_shared.plugins.plugin_manager import hook_implementation
-    from ert_shared.plugins.plugin_response import plugin_response
+    from ert.shared.plugins.plugin_manager import hook_implementation
+    from ert.shared.plugins.plugin_response import plugin_response
 except ModuleNotFoundError:
```
berland, line 9: (this will mask errors)

berland:
You need to pin `ert>=2.38.0b5` in `setup.py` in order to be able to fetch a version from pypi with `ert.shared`.

ecl2df/hook_implementations/jobs.py
asnyv: ah, yes of course

asnyv:
Know we would have to pin, therefore I was a bit surprised that it based after my quick search and replace all @berland :wink: but wasn't sure if we wanted to pin a minimum requirement prerelease, or just wait for an official ert release?

berland:
If you merge without pinning, the ert plugins from ecl2df will stop working (for those cloning ec2ldf) until ert 2.38 is released.

setup.py
```diff
@@ -36,7 +36,11 @@
     "sphinx-autodoc-typehints",
     "sphinx_rtd_theme",
 ]
-EXTRAS_REQUIRE = {"tests": TEST_REQUIREMENTS, "docs": DOCS_REQUIREMENTS, "ert": ["ert"]}
+EXTRAS_REQUIRE = {
+    "tests": TEST_REQUIREMENTS,
+    "docs": DOCS_REQUIREMENTS,
+    "ert": ["ert>=2.38.0-b5"],
```
berland, line 42: one dash too many: https://pypi.org/project/ert/2.38.0b5/

berland (approved):

anders-kiaer:
A bit late to the party here, but if I remember correctly, having a prerelease lower bound in `setup.py` will actually tell `pip` to always use the latest (pre)release of that `package`). Probably not a problem since it is an extra `ecl2df[ert]`, but we  might want to change the lower bound to official release `2.38` in `ecl2df` when `ert` is released.

asnyv:
@anders-kiaer that is the plan 👍

Issue #399, opened 2022-02-09
Tests fail with Pandas 1.4.0


berland:
`test_parameters::test_load` will fail with Pandas 1.4.0, as that Pandas version seems quite ok with parsing text files with more fields than what it has column names for. Pandas <1.4.0 gives a ParserError.

Solutions: 
1. Pin Pandas < 1.4.0
2. Make `pd.read_csv()` in `ecl2df.parameters.load_parameterstxt` behave as strictly as needed.
3. Relax the testing, not try to capture all error scenarios with parameters.* this early.

asnyv:
Closed by #411

PR #414, opened 2022-08-19
Stricten parsing of malformed parameter csv's with pandas>=1.4

### LINKED ISSUE #411: Relax pandas old version requirement
This should be relaxed:

https://github.com/equinor/ecl2df/blob/1cd9d887c75fb171100d2bbc70ad2cb02f9ddbdd/setup.py#L22

asnyv:
Resolves #411

### DIFF
--- a/ecl2df/parameters.py
+++ b/ecl2df/parameters.py
@@ -3,6 +3,7 @@ related to Eclipse runs"""
 
 import json
 import logging
+import warnings
 from pathlib import Path
 from typing import Any, Dict, List, Union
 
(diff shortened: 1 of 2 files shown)


asnyv:
Used `index_col=False` to avoid the `pandas>1.4` behavior where 3 columns in the data and 2 defined `names` results in the first column being used as index. 
Then catching the warning `pandas` writes when some data is dropped (column 3) and raise it as an error instead.

asnyv:
also closes #399

ecl2df/parameters.py
```diff
@@ -65,9 +66,20 @@ def load_parameterstxt(filename: Union[str, Path]) -> Dict[str, Any]:
         filename: file containing one key-value pair pr. line,
             separated by whitespace
     """
-    dframe = pd.read_csv(
-        filename, comment="#", sep=r"\s", engine="python", names=["KEY", "VALUE"]
-    )
+    with warnings.catch_warnings(record=True):
+        warnings.filterwarnings("error")
```
berland, line 70: Maybe a small comment here is warranted for the future developer?

asnyv: done

berland (approved):
👍

PR #420, opened 2022-09-26
Breaking: `ecl2csv summary.df` arg `include_restart` default changed to `False`

### LINKED ISSUE #419: `ecl2csv summary` should expose `include_restart`
Currently it is not possible to set `include_restart=False` when using `ecl2csv summary`. This creates a lot of overhead, as users often will have many restarts based on the same base run, and adding the same base data to all restarted ensembles is unnecessary in most cases. 

The ideal solution would in my opinion be that you have to **add** the flag `--include_restart` to get the restart data, but that would be a breaking change... An alternative is to add a flag like `--skip_restart`, but I fear that the many users will then continue to get way more data than they really need, as they will not be aware of the possibility to drop it.

asnyv:
Resolves #419 

This is a suggestion. It is a breaking change. On the other hand, the data from prior to restart is only included today if the path is less than a certain number of characters, so it is not completely stable. Therefore think it might be better to have a stable default and rather make the user able to request restart data if necessary.

### DIFF
--- a/.github/workflows/ecl2df.yml
+++ b/.github/workflows/ecl2df.yml
@@ -12,13 +12,12 @@ on:
       - published
   schedule:
     # Run CI every night and check that tests are working with latest dependencies
-    - cron:  '0 0 * * *'
+    - cron: '0 0 * * *'
 
 env:
   ERT_SHOW_BACKTRACE: 1
 
 jobs:
-
   ecl2df:
     runs-on: ubuntu-18.04
     strategy:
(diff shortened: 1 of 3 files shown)


berland (approved):
ok

asnyv:
@berland Thinking a bit, should we also change the default for the `summary.df` function's `include_restart` to `False`? This is of course even more breaking as it is then propagating to repositories that depend on `ecl2df`

PR #421, opened 2022-09-26
Set python 3.10.5 back to 3.10

lindjoha:
Python 3.10 version was set to 3.10.5 because of a `mypy` error. This should be fixed in the next version of `mypy` and then this PR can be merged.


asnyv (approved):
Rebase før merge

Issue #423, opened 2022-10-13
[summary] Missing header name for index updated with timestep


alifbe:
After PR #412 we can see that the date index precision is improved. But the header name is not retained. This caused issue with webviz plugin as the CSV generated doesn't have proper header name

![image](https://user-images.githubusercontent.com/11570927/195602729-efab5007-c776-4564-857c-304916a0afe3.png)

alifbe:
The issue is resolved with #424

PR #424, opened 2022-10-13
[summary] FIX: Ensure index name is retained

alifbe:
This PR fixes issue #423 where data frame index header name is set to None when the index is replace with list. It makes the generated CSV file unusable for WebViz since WebViz cannot fine the column.

### DIFF
--- a/tests/test_summary.py
+++ b/tests/test_summary.py
@@ -804,7 +804,7 @@ def test_unique_datetime_for_short_timesteps(filepath):
 )
 def test_unique_datetime_retain_index_name(filepath):
     """Test _ensure_unique_datetime_index method retain index name"""
-    assert summary.df(EclFiles(filepath)).index.name is not None
+    assert summary.df(EclFiles(filepath)).index.name == "DATE"
 
 
 def test_smry_meta():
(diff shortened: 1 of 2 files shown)


alifbe:
@asnyv can you please review this small PR?

tests/test_summary.py
```diff
@@ -795,6 +795,18 @@ def test_unique_datetime_for_short_timesteps(filepath):
...
+@pytest.mark.parametrize(
+    "filepath",
+    [
+        SHORT_STEP_WITH_TIMESTEP,
+        SHORT_STEP_WITH_TIMESTEP_LONG,
+    ],
+)
+def test_unique_datetime_retain_index_name(filepath):
+    """Test _ensure_unique_datetime_index method retain index name"""
+    assert summary.df(EclFiles(filepath)).index.name is not None
```
asnyv, line 807: The index should be names DATE in all scenarios I think, so maybe better to check that than not None?
```suggestion
    assert summary.df(EclFiles(filepath)).index.name == "DATE"
```

alifbe: I see. I was not sure if it's always DATE or it could be TIME for example. 

I'll update it :+1:

asnyv (approved):
Looks good, thanks 😊 remember to select squash commits when merging

PR #441, opened 2023-05-02
Bulk export

daniel-sol:
Add functionality for exporting with metadata.
This makes upload of data to cloud solutions such as sumo straight forward


daniel-sol:
Will squash and merge, then try again :-)

PR #442, opened 2023-05-08
Bulk export

daniel-sol:
Bulk export of data with metadata


ecl2df/common.py
```diff
@@ -114,13 +115,140 @@
...
+    try:
+        config = yaml_load(args["config_path"])
+    except TypeError:
+        logger.debug("Config allready a dictionary, no need to read it")
+        config = args
+    logger.debug(config.keys())
+    exp = ExportData(
+        config=config,
+        name=name,
+        tagname=re.sub(r"\..*", "", tagname),
```
perolavsvendsen, line 233: Consider separating into small function, and explicitly test.

ecl2df/common.py
```diff
@@ -114,13 +115,140 @@
 logger: logging.Logger = logging.getLogger(__name__)
 
 
-def write_dframe_stdout_file(
+def set_name_from_args(args):
```
berland, line 118: I don't quite get what this is doing from reading the function name in the context of common.py. Can it get a more descriptive name, optionally an example in the docstring?

ecl2df/bulk.py
```diff
@@ -0,0 +1,197 @@
...
+        parser (argparse.ArgumentParser or argparse.subparser): parser to
+            fill with arguments
+    """
+    # parser.add_argument("DATAFILE", help="Name of Eclipse DATA file.")
+    parser.add_argument("-v", "--verbose", action="store_true", help="Be verbose")
+    return parser
+
+
+def bulk_export(eclpath, config_path, include: List = None, options: dict = None):
+    """Bulk uploads every module to sumo with metadata
```
berland, line 66: where does sumo come into play in this function? Is this functionality only available with sumo?

ecl2df/compdat.py
```diff
@@ -955,8 +955,12 @@ def fill_parser(parser: argparse.ArgumentParser) -> argparse.ArgumentParser:
...
         "--output",
         type=str,
-        help="Name of output csv file.",
-        default="compdat.csv",
+        help=(
+            "Override name of output csv file.\n"
+            + "Otherwise name is derived from datafile and datatype.\n"
+            + "Use '-' for stdout."
+        ),
+        default=None,
```
berland, line 963: API breaking change?

ecl2df/common.py
```diff
@@ -114,13 +115,142 @@
...
+
+    name = re.sub(r"(-\d+)?\..*", "", base_name)
+    logger.debug("Name and tag " + name + "|" + tagname)
+
+    return name, tagname, content
+
+
+def write_dframe_to_file(
     dframe: Union[pd.DataFrame, pyarrow.Table],
     output: str,
```
berland, line 182: I don't like changing this arg from a well-defined string to a big and complex object (an argparse.Namespace), making it enormously harder to test. Can this be refactored?

ecl2df/faults.py
```diff
@@ -92,7 +113,7 @@ def faults_main(args) -> None:
     faults_df = df(deck)
     write_dframe_stdout_file(
         faults_df,
-        args.output,
+        args,
```
berland, line 116: The function should not be given all of the args, you only want to provide the absolute minimum number of inputs it needs.

berland (changes requested):
Not having been involved earlier, but summoned to review I have some questions:

* What is the issue? (as an issue in ecl2df isse tracker)
* Why does this functionality belong inside the ecl2df repository (there are probably reasons, I am asking control questions)
* Why can't this be a shell script outside of ecl2df (or in a different repo as python code)
* ecl2df, by its name, is about converting from/to eclipse data to/from dataframes. Is this PR within that scope, or is it stretching it (by being glue code from dataframes to other systems?) (I know, there is code already in ecl2df that is stretching this original intention).
* This PR adds more 5000 lines, which is very significant in terms of lines already existing. Can the added features from this PR defend that many lines, giving the associated extra future maintenance burden?
* There looks to be API breaks in the PR. This should be sorted out, preferably in separate PRs. If ecl2df needs to be changed in order to facilitate the PR, having these changes as "preporatory" PRs would be nice, making each change in the git history understandable later.
* There is already "metadata" support in ecl2df, as in https://github.com/equinor/ecl2df/blob/744394994eafb60014d6b7436c39339cf8894957/ecl2df/summary.py#L418 f.ex. How does this relate to the new "metadata" stuff?

berland (changes requested):
Not having been involved earlier, but summoned to review I have some questions:

* What is the issue? (as an issue in ecl2df isse tracker)
* Why does this functionality belong inside the ecl2df repository (there are probably reasons, I am asking control questions)
* Why can't this be a shell script outside of ecl2df (or in a different repo as python code)
* ecl2df, by its name, is about converting from/to eclipse data to/from dataframes. Is this PR within that scope, or is it stretching it (by being glue code from dataframes to other systems?) (I know, there is code already in ecl2df that is stretching this original intention).
* This PR adds more 5000 lines, which is very significant in terms of lines already existing. Can the added features from this PR defend that many lines, giving the associated extra future maintenance burden?
* There looks to be API breaks in the PR. This should be sorted out, preferably in separate PRs. If ecl2df needs to be changed in order to facilitate the PR, having these changes as "preporatory" PRs would be nice, making each change in the git history understandable later.
* There is already "metadata" support in ecl2df, as in https://github.com/equinor/ecl2df/blob/744394994eafb60014d6b7436c39339cf8894957/ecl2df/summary.py#L418 f.ex. How does this relate to the new "metadata" stuff?

daniel-sol:
Will split into more managable PR's

PR #444, opened 2023-05-12
Bump versions in CI

berland:


xjules (approved):
Was about to do it :) :rocket:

Issue #445, opened 2023-05-22
Auto detect name of output file from name of datafile and submodule name (datatype)


daniel-sol:
Make package auto detect name of output from the various submodules in stead of this being a user defined input

* Suggestion is to have the name as "name of datafile"--"name of submodule". csv/.arrow
* It is suggested to still keep output as user input, but default to None, with auto detection when it is None, this enables user to still set name if desired.

daniel-sol:
It is decided that for the time being this will not be implemented.

PR #447, opened 2023-05-30
Autodetect name

daniel-sol:
Answering issue #445. Is part of answers to review of closed pull  #442 
* Main point is autodetection of name for output from all submodules
* Autodetected name is <name of the eclipse/opm datafile - number at end>--<name of submodule>
* Rationale for subtracting number at end is that this number is most commonly a number applied automatically when running an ensemble
  (And is therefore more logical when collating results from an ensemble)
* Name can still be overwritten with the argument output, but is defaulted to None, which then triggers autodetection
* Otherwise aligned with latest commits (as of 2023-05-30)


ecl2df/compdat.py
```diff
@@ -955,8 +955,12 @@ def fill_parser(parser: argparse.ArgumentParser) -> argparse.ArgumentParser:
...
         "--output",
         type=str,
-        help="Name of output csv file.",
-        default="compdat.csv",
+        help=(
+            "Override name of output csv file.\n"
+            + "Otherwise name is derived from datafile and datatype.\n"
+            + "Use '-' for stdout."
+        ),
+        default=None,
```
perolavsvendsen, line 963: This changes the default behavior? Will this create problems for existing users (who will expect a file called `compdat.csv` still?)

perolavsvendsen: Is it better to explicitly trigger the auto detection (at least for a period of time), and look into changing the defaults later when we have confirmed that it is stable? Could be quite a lot of unexpected effects here on existing usage. Perhaps using `name=auto` or similar. That way, the feature is included and available, but it still gives a chance to learn how it works "in the wild".

daniel-sol: Well, you of course have a point when it comes to existing users, but my feeling is that this is rarely executed with only the default. Because this will give the user a file names compdat.csv in the current folder

 What is implemented now is parallel to this but makes a file called <name of datafile>--compdat.csv in stead. Still in the current folder.. 

If they use --output the behaviour will be identical to before. But what I am more worried about is if people want autodetection, but put also to put it in a different folder, that will not work. So I have considered to make functionality that check if output is a directory, and if so prepend this to output name. Remember that most of the time this will be run through ert as a forward model called ECL2CSV, here I see that you need to put in <output>. About the autodetection as a specific option, to me this is basically what None does. But it would be good to get feedback from people that use this more regularly..

daniel-sol:
Closed to rebase, then making new

PR #451, opened 2023-06-15
Autodetect name

daniel-sol:
Version two of #447 after rebase to fix CI issues. Please refer to comments under previous PR

Answering issue https://github.com/equinor/ecl2df/issues/445. Is part of answers to review of closed pull https://github.com/equinor/ecl2df/pull/442

Main point is autodetection of name for output from all submodules
Autodetected name is <name of the eclipse/opm datafile - number at end>--
Rationale for subtracting number at end is that this number is most commonly a number applied automatically when running an ensemble
(And is therefore more logical when collating results from an ensemble)
Name can still be overwritten with the argument output, but is defaulted to None, which then triggers autodetection
Otherwise aligned with latest commits (as of 2023-06-22)

### DIFF
--- a/ecl2df/fipreports.py
+++ b/ecl2df/fipreports.py
@@ -202,7 +202,15 @@ def fill_parser(parser: argparse.ArgumentParser) -> argparse.ArgumentParser:
         default="FIPNUM",
     )
     parser.add_argument(
-        "-o", "--output", type=str, help="Output CSV filename", default="outflow.csv"
+        "-o",
+        "--output",
+        type=str,
+        help=(
+            "Override name of output csv file.\n"
+            + "Otherwise name is derived from datafile and datatype.\n"
+            + "Use '-' for stdout."
+        ),
+        default=None,
     )
     parser.add_argument("-v", "--verbose", action="store_true", help="Be verbose")
     parser.add_argument("--debug", action="store_true", help="Debug mode for logging")
@@ -219,4 +227,4 @@ def fipreports_main(args) -> None:
     else:
         prtfile = EclFiles(args.PRTFILE).get_prtfilename()
     dframe = df(prtfile, args.fipname)
-    write_dframe_stdout_file(dframe, args.output, index=False, caller_logger=logger)
+    write_dframe_stdout_file(dframe, vars(args), index=False, caller_logger=logger)
--- a/ecl2df/vfp/_vfp.py
+++ b/ecl2df/vfp/_vfp.py
@@ -453,8 +453,13 @@ def fill_parser(parser: argparse.ArgumentParser) -> argparse.ArgumentParser:
         "-o",
         "--output",
         type=str,
-        help="Name of output csv file. No CSV dump if empty",
-        default="",
+        help=(
+            "Override name of output csv file.\n"
+            + "Otherwise name is derived from datafile and datatype.\n"
+            + "Use '-' for stdout."
+            + "Note!! No csv dump if empty"
+        ),
+        default=None,
     )
     parser.add_argument(
         "-k",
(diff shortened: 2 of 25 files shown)


ecl2df/vfp/_vfp.py
```diff
@@ -506,9 +511,9 @@ def vfp_main(args) -> None:
             table_number = int(
                 vfp_table.schema.metadata[b"TABLE_NUMBER"].decode("utf-8")
             )
-            vfp_filename = f"{outputfile}_{str(table_number)}.arrow"
+            args["output"] = f"{outputfile}_{str(table_number)}.arrow"
```
perolavsvendsen, line 514: Confusing for later readers that things are added to the variables (?) outside the scope of giving variables?

tests/test_gruptree.py
```diff
@@ -435,14 +435,6 @@ def test_emptytree_commandlinetool(tmp_path, mocker, caplog):
     assert "No tree data to prettyprint" in caplog.text
 
 
-def test_cli_nothing_to_do(mocker, capsys):
```
perolavsvendsen, line 438: Removal of this test indicates some change in behavior, in this case "when DATA is supplied, but no action". Do we know if this is acceptable (or disruptive) for current usage?

tests/test_setting_name.py
```diff
@@ -0,0 +1,106 @@
...
+
+    correct_name = "2_R001_REEK"
+    correct_tag = "mycommand"
+    name, tag = find_name_components(args)
+    assert name == correct_name, f"Name is {name}, should be {correct_name}"
+    assert tag == correct_tag, f"Tag is {tag}, should be {correct_tag}"
+
+
+def test_set_name_from_args(name_args):
+    """Test function set_name_from_args
```
perolavsvendsen, line 74: Seems to not do this, it is testing `make_output_name`. Should the test name and the docstring reflect better what it is testing?

tests/test_setting_name.py
```diff
@@ -0,0 +1,106 @@
...
+            submod_name,
+            REEK,
+        ],
+    )
+    ecl2csv.main()
+    outpath = tmp_path.glob("*.csv")
+    file_names = list(outpath)
+    assert len(file_names) == 1, "Produced more than one file"
+    file_name = file_names[0]
+    print(f" file name in test {file_name}")
```
perolavsvendsen, line 107: Remove `print` statements?

perolavsvendsen (commented):
Others who knows this package better and can better see consequences must do a more proper review, but I have left some comments/suggestions. Also, I suggest expanding the tests a little bit to include more "weird" things and corner cases for the autodetect functionality. How does it handle special characters (should it?), am I allowed to produce illegal filenames etc.

tests/test_setting_name.py
daniel-sol: Well spotted, renamed function and forgot to rename in docstring.. Will fix..

tests/test_setting_name.py
daniel-sol: Will do

ecl2df/fipreports.py
```diff
@@ -202,7 +202,15 @@ def fill_parser(parser: argparse.ArgumentParser) -> argparse.ArgumentParser:
...
-        "-o", "--output", type=str, help="Output CSV filename", default="outflow.csv"
+        "-o",
+        "--output",
+        type=str,
+        help=(
+            "Override name of output csv file.\n"
+            + "Otherwise name is derived from datafile and datatype.\n"
+            + "Use '-' for stdout."
+        ),
+        default=None,
```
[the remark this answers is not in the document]
perolavsvendsen: Yes, so a combination of deterministic folder but autodetected name. Sounds like new functionality, e.g "let user decide the folder". Suggest making an issue, agree on the specs then a separate PR later.

tests/test_gruptree.py
daniel-sol: All it used to do was check if sys.exit is called if there are no results to produce, but one could off course bring it back, but change the test..

berland:
What is the issue solved by this PR? (please write an issue describing the problem, and link it)

daniel-sol:
After meeting 17/8-2023 it was decided to not do this implementation in ecl2df directly because of the dependencies it would introduce. At the time of writing it seems likely that the conceptual ideas behind it will live on in fmu-sumo as a utlitity there.

PR #454, opened 2023-10-09
Use all accessible ports for ERT on local queue

berland:
If not, ERT will pick a port in the default range 51820-51840 for which should be saved for ERT runs over the network. When running tests in parallel, using the limited range exposes the tests to port starvation.


berland:
Superseded by https://github.com/equinor/ert/pull/6280

PR #456, opened 2023-10-11
Remove use of pkg_resource

### LINKED ISSUE #455: pkg_resources is deprecated
Every invocation of ERT (using parts of ecl2df or not) will print the warning to the users console:

```
/prog/komodo/bleeding-2023101106-py38-rhel7/root/lib/python3.8/site-packages/ecl2df/hook_implementations/jobs.py:4: DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html
  from pkg_resources import resource_filename
```

This should be fixed to avoid uneccesary noise.

JHolba:
This import is causing warning messages in all programs that import this library.

fixes #455

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -14,8 +14,8 @@
 # import os
 # import sys
 # sys.path.insert(0, os.path.abspath('.'))
-import pkg_resources
 import datetime
+from importlib import metadata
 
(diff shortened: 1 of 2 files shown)


berland (approved):
good to go!

PR #457, opened 2023-10-11
Pin pydantic

JHolba:
newest pydantic is pulled in by rstcheck. ert requires pydantic<2

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -31,6 +31,7 @@ DOCS_REQUIREMENTS = [
     "autoapi",
     "ipython",
     "rstcheck",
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -23,6 +23,7 @@
     "pyarrow",
     "pyyaml>=5.1",
     "treelib",
+    "pydantic<2",
```
berland, line 26: Since this is an indirect dependency, could you add a comment on when it can be removed again?

JHolba: updated with comment

berland (approved):

PR #461, opened 2023-11-14
Add support to pandas 2

alifbe:
Add support to pandas 2 and add pandas version to test matrix


tests/test_summary.py
```diff
@@ -942,7 +942,10 @@ def test_smry_meta_synthetic():
 def test_fix_dframe_for_libecl(dframe, expected_dframe):
     """Test the dataframe preprocessor/validator for df2eclsum works"""
     pd.testing.assert_frame_equal(
-        _fix_dframe_for_libecl(dframe), expected_dframe, check_index_type=False
+        _fix_dframe_for_libecl(dframe),
+        expected_dframe,
+        check_index_type=False,
+        check_column_type=False,
```
alifbe, line 948: We need to set `check_column_type=False` because in Pandas > 2.0 set empty data frame to RangeIndex and that's not the case when you delete a column. Hence it will fail on case 10-12

tests/test_summary.py
```diff
@@ -1115,13 +1118,6 @@ def test_df2pyarrow_500years():
     dframe.index.name = "BOGUS"
     pyat = _df2pyarrow(dframe)
 
-    # pylint: disable=c-extension-no-member
-    with pytest.raises(pyarrow.lib.ArrowInvalid):
-        # We cannot convert this back to Pandas, since it will bail on failing
-        # to use nanosecond timestamps in the dataframe object for these dates.
-        # This is maybe a PyArrow bug/limitation that we must be aware of.
-        pyat.to_pandas()
-
```
alifbe, line 1124: This exception doesn't seem to happen anymore.

ecl2df/wellcompletiondata.py
```diff
@@ -228,7 +228,7 @@ def _df2pyarrow(dframe: pd.DataFrame) -> pyarrow.Table:
         else:
             field_metadata = {}
         if colname == "DATE":
-            dtype = pyarrow.timestamp("ms")
+            dtype = pyarrow.timestamp("ns")
```
berland, line 231: Careful with this, with nanoseconds you will hit a limit in year 2262, and summary files going past 2262 is something we need. Why is this needed for pandas2?

alifbe: Good point. 👍 

From what I can see, pyarrow 12.0.0 `to_pandas()` method will convert `DATE` column to `datetime64[ns]` despite what we set on the line that you point above. The new pyarrow 14.0.0 does a proper conversion and causing a `dtype` issue in the test below since pandas by default will set dtype to `datetime64[ns]`. I think we can just get away with it by specifying `check_dtype=False` 

https://github.com/equinor/ecl2df/blob/798ee6225697919cad4876d2390e3aa209ba48e0/tests/test_wellcompletiondata.py#L89

Should we ensure that all date column from ecl2df will be using `datetime64[ms]` ?

berland: There are tests regarding year 2262 support, check that these cover this case (export to arrow format)

berland (approved):
LGTM. Ensure squashing when merging.

PR #470, opened 2023-11-28
Make things a bit less circular

yngve-sk:


jonathan-eq (approved):
Looks good! Good job!

PR #472, opened 2023-12-04
Expect RuntimeError or ValueError in ntequl test

mferrera:
Relates to Python 3.11+ support

Later versions of opm use a newer pybind11 version and potentially different exceptions for input errors in its C++ backend. This makes both `RuntimeError` and `ValueError` possible but tricky to pin down to a particular opm version. This takes the easy route and allows for both.


berland (approved):
:+1:

mferrera:
I can't merge this myself -- #471 is also for Python 3.11 😄

PR #474, opened 2023-12-07
Ensure some actions are trigger only for pandas>2

alifbe:
In PR #461 I have added `pandas-version` to the test matrix but that causes issue in the github action `Update GitHub pages` and potentially `Build python package and publish to pypi`

Add condition: `matrix.pandas-version == 'pandas>2'` to the action


berland (approved):
👍🏻

Issue #475, opened 2023-12-18
Pydantic V2


pinkwah:
Support Pydantic v2 as it's planned for January 2024 komodo release.

Currently blocked by ERT and Semeio.

PR #476, opened 2023-12-21
Remove Pydantic as dependency

pinkwah:
Resolves: #475

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -31,7 +31,6 @@ DOCS_REQUIREMENTS = [
     "autoapi",
     "ipython",
     "rstcheck",
-    "pydantic<2",  # remove this when ert unpins fastapi and pydantic
     "sphinx<7",
     "sphinx-argparse",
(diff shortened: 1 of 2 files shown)


TeeeJay:
Sorry to bother, but would it make sense to add <3 or ^2 as a dependency resolver instead? That is what we in eCalc ...

pinkwah:
I am not 100% sure about this, but I think `^2` is a Poetry thing and not Python. You are correct that it would be better to have an upper limit, but maintaining this without Poetry would be too difficult.

setup.py
```diff
@@ -31,7 +31,7 @@
     "autoapi",
     "ipython",
     "rstcheck",
-    "pydantic<2",  # remove this when ert unpins fastapi and pydantic
+    "pydantic",
```
berland, line 34: Remove the entire line, as pydantic is a transitive dependency only.

berland (approved):
👍🏻

PR #479, opened 2024-01-26
Test summary extraction through RESTART

berland:
This acts as a regression test for https://github.com/equinor/resdata/pull/948


eivindjahren (approved):
Nice! I have a similar setup here: https://github.com/equinor/xtgeo/blob/main/tests/test_opm_integration/test_grid_io.py which uses opm.

PR #485, opened 2024-02-07
Parametrize length of DATA file in summary restart test

berland:
Also give up testing when the pytest tmp dir is too long for Eclipse's 132 chars per line limitation


jonathan-eq (approved):
LGTM!

PR #496, opened 2024-08-13
Run tests also for python 3.11-12

larsevj:
This ended up a bit bigger than expected. One downside with separating out style and types in their own workflows is that they are most likely only tested with pandas>=2.
A bit unsure how to split/order the commits.

### DIFF
--- a/ci/testkomodo.sh
+++ b/ci/testkomodo.sh
@@ -2,10 +2,6 @@ install_test_dependencies () {
   pip install -r test_requirements.txt
 }
 
-install_package () {
-  pip install .[tests,docs]
-}
-
 start_tests () {
   pytest --run-eclipse-simulator
 }


README.md
```diff
@@ -1,7 +1,7 @@
 [![Build Status](https://img.shields.io/github/workflow/status/equinor/res2df/res2df)](https://github.com/equinor/res2df/actions?query=workflow%3Ares2df)
 [![codecov](https://codecov.io/gh/equinor/res2df/graph/badge.svg?token=3sZBGGu5VG)](https://codecov.io/gh/equinor/res2df)
-[![Python 3.8-3.10](https://img.shields.io/badge/python-3.8%20|%203.9%20|%203.10-blue.svg)](https://www.python.org)
-[![Code style: black](https://img.shields.io/badge/code%20style-black-000000.svg)](https://black.readthedocs.io/)
+[![Python 3.8-3.12](https://img.shields.io/badge/python-3.8%20|%203.9%20|%203.10-blue.svg)](https://www.python.org)
```
berland, line 3: The URL here must be updated to 3.12 as well.

berland:
Maybe merge commit nr 2 and 4, since the repo at the state of commit 2 might not pass the pre-commit tests?

berland (approved):
Great!

PR #502, opened 2024-09-09
Unpin sphinx

berland:


berland:
Looks like sphinx8 needs pyproject.toml based setup (?)

PR #509, opened 2024-11-25
Avoid deep copying of the attrs when making arrow

berland:
Pandas 2.2 has changed behaviour in treating the experimental 'attrs', and will copy the entire dict (for all columns) each time dframe[colname] is called.

Instead of calling dframe[colname] for every column, we prepare the values upfront instead.


berland:
As a bonus, the code is now also 25% faster also on Pandas 2.1.4

eivindjahren (approved):

PR #512, opened 2024-11-27
Use resfo for parsing headers in UNRST file

berland:
Using resdata to ask for headers implies reading the entire file which requires excessive memory.

Porting more of res2df to use resfo is postponed to a future issue.

Before this commit, obtaining the RST information for one particular date in a 118GB UNRST file was not possible (requires at least more than 118GB of available memory). With this commit, the same operation can be done in arund 6 GB of used memory.

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -3,6 +3,9 @@
 [mypy-resdata.*]
 ignore_missing_imports = True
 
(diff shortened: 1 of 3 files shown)


eivindjahren (approved):
Amazing!

PR #514, opened 2024-11-28
Avoid crash-causing xtgeo version in Python 3.8

mferrera:
xtgeo 4.3.2 caused crashes with pandas>2. Versions greater than this have resolved the problem, so far.


mferrera:
Not sure if you prefer to just unpin or avoid it entirely, but happy to remove the entry as well.

PR #521, opened 2025-01-17
Solve Pandas 2.2 warnings

berland:


eivindjahren (approved):
Could consider running tests with -WError to catch new deprecation warnings in the future, but not required.

PR #526, opened 2025-02-12
Make sourcing of eclrun non-flaky

berland:
If there is a simple bash test that fails in the sourced file, the sourcing will fail when it is run in an environment where 'set -e' is set. Ensure this is temporarily unset


andreas-el (approved):
🏖️

PR #529, opened 2025-04-11
Solve and enforce ruff E rules

berland:


berland:
Needs #532 to pass tests.

asnyv (approved):
🐣

PR #532, opened 2025-04-14
Remove sphinx configuration no longer valid

berland:
Resolves CI build failure in https://github.com/equinor/res2df/actions/runs/14398733428/job/40379506011

Removing these settings does not affect the built documentation as far as I can see.

Building sphinx on a setup.py project is deprecated.


andreas-el (approved):
🧹 🧹 💥

PR #533, opened 2025-04-14
Use Erts newer plugin system for forward model steps

berland:
Solves #531


andreas-el (approved):
Nicely done! 🏖️

res2df/hook_implementations/forward_model_steps.py
```diff
@@ -0,0 +1,138 @@
...
+
+        return decorator
+
+    class ForwardModelStepPlugin:  # type: ignore
+        pass
+
+    class ForwardModelStepJSON:  # type: ignore
+        pass
+
+    class ForwardModelStepDocumentation:  # type: ignore
```
andreas-el, line 25: This could have been swapped with`ForwardModelStepPlugin` to get perfect ordering

berland: we like order

PR #536, opened 2025-04-22
res2csv summary - do not create output when summary file doesn't exist

### LINKED ISSUE #503: `res2csv summary` - do not create an arrow file when the summary file does not exist
If one run `res2csv summary --verbose --output foo.arrow --time_index raw --arrow -- OOF`, and `OOF.UNSMRY` does not exist, the `foo.arrow` file will still be created. Would it be better to not create an output file in this situation?

```
> res2csv summary --verbose --output foo.arrow --time_index raw --arrow -- OOF
WARNING:res2df.summary:Error reading summary instance, returning empty dataframe
INFO:res2df.summary:Writing to file foo.arrow
```

alifbe:
Close #503 

Do not create output file when summary file doesn't exist or invalid. Create a unit test for it as well.


tests/test_summary.py
```diff
@@ -180,6 +180,32 @@ def test_res2csv_summary(tmp_path, mocker):
     assert str(disk_df["DATE"].values[-1]) == "2003-01-02"
 
 
+@pytest.mark.integration
+def test_res2csv_summary_no_output_on_invalid_input(tmp_path, mocker):
```
berland, line 184: This tests only tests for missing input, not invalid. That should be reflected in the test name (and then the comment below becomes superfluous)

alifbe: You are right. I have now added another test with empty input file to simulate invalid input.

berland: Good. I would have split it into two test functions as well.

berland (approved):
(remember to squash)

tests/test_summary.py
```diff
@@ -180,6 +180,58 @@ def test_res2csv_summary(tmp_path, mocker):
...
 
 
+@pytest.mark.integration
+def test_res2csv_summary_no_output_on_missing_and_invalid_input(tmp_path, mocker):
+    """Test that no output file is created when the input file is missing or invalid."""
+
+    tmp_input_file = tmp_path / "MISSING_INPUT.UNSMRY"
+    tmp_output_file = tmp_path / "missing_output.arrow"
+
+    tmp_input_file.touch()
```
berland, line 190: (this can be confusing later, the filename indicates it is missing, but then it is not after you have touched it)

tests/test_summary.py
```diff
@@ -180,6 +180,58 @@ def test_res2csv_summary(tmp_path, mocker):
...
+    tmp_output_file = tmp_path / "invalid_output.arrow"
+
+    # Mock the command-line arguments to simulate invalid input
+    mocker.patch(
+        "sys.argv",
+        [
+            "res2csv",
+            "summary",
+            "--verbose",
+            "INVALID_INPUT.UNSMRY",
```
berland, line 221: ah, this is your "missing" file?

Easier to get this right by splitting into two test function to avoid relying on which files are in tmpdir.

berland:
Great!

PR #541, opened 2025-05-08
Pin snowballstemmer version

alifbe:
Snowballstemmer v3.0 released a few hours ago is causing issue with Sphinx. https://github.com/snowballstem/snowball/issues/229

Pin it for now.


rnyb (approved):
Consider following up Copilot suggestion to add inline comment/context

alifbe:
Not needed anymore

PR #542, opened 2025-05-09
res2df maintenance

### LINKED ISSUE #469: Modernize build system
By the looks it would not be too much work to move to [a PEP518/621 build system](https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/) based on `pyproject.toml`.

### LINKED ISSUE #501: Relax sphinx requirement
This needs to be updated, as we need to support sphinx 8

https://github.com/equinor/res2df/blob/a52e74e07150afd5d1e1c88404308396b81fe678/setup.py#L40

### LINKED ISSUE #538: Update setup configuration
Consider modernize setup configuration (replace setup.py with pyproject.toml)

alifbe:
Close #469
Close #538 
Close #501 
A few commits:
1. Move code from `res2df `to `src/res2df`
2. Moving from setup.py to pyproject.toml 
3. Activate nightly test
4. Relax sphinx requirement


berland:
`version.py` is tracked in the first commit here, it should not.

src/res2df/version.py
```diff
@@ -0,0 +1,21 @@
+# file generated by setuptools-scm
+# don't change, don't track in version control
```
berland:
```
# file generated by setuptools-scm
# don't change, don't track in version control
```

alifbe: Thanks, rebased the commits now.

berland (approved):
👍🏻

PR #546, opened 2025-08-05
Fix build issue after moving from setup.py to pyproject.toml

alifbe:
Fix build issue after migrating from setup.py to pyproject.toml


rnyb (approved):
👍

PR #550, opened 2025-09-08
Make sure PORV column is only processed once in init2df

### LINKED ISSUE #517: egrid reading dimesionality error / BUG
Image

I'm getting this error when trying to read an egrid file, however checking the error seems to be a dimensionality error in the grid.py, more specifically the function init2df, this line 
`init_df["PORV"] = porv_numpy[glob_idxs].reshape(-1, 1)`
should be changed by 
`init_df["PORV"] = porv_numpy[glob_idxs].reshape(-1)`
after I made the change in my local version the script works
Image

larsevj:
Resolves #517 

Difficult to say if this pr resolves the exact same issue that is reported since the information is quite sparse, but the tests added in this pr reproduce the exact same error message, which happens when one tries to add the PORO column of shape (-1,1) to an empty pandas dataframe. Reshaping it should not be necessary since pandas expect a 1d numpy array.

### DIFF
--- a/src/res2df/grid.py
+++ b/src/res2df/grid.py
@@ -438,13 +438,12 @@ def init2df(
     usevectors = []
     include_porv = False
     for vec in init.headers:
-        if vec[1] == egrid.getNumActive() and any(
+        if vec[0] == "PORV" and any(fnmatch.fnmatch("PORV", key) for key in vectors):
+            include_porv = True
+        elif vec[1] == egrid.getNumActive() and any(
             fnmatch.fnmatch(vec[0], key) for key in vectors
         ):
             usevectors.append(vec[0])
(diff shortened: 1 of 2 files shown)


src/res2df/grid.py
```diff
@@ -438,13 +438,12 @@ def init2df(
     usevectors = []
     include_porv = False
     for vec in init.headers:
-        if vec[1] == egrid.getNumActive() and any(
+        if vec[0] == "PORV" and any(fnmatch.fnmatch("PORV", key) for key in vectors):
+            include_porv = True
+        elif vec[1] == egrid.getNumActive() and any(
```
alifbe, line 443: Good point. 👍 This will ensure that PORV will only be process once when all cells are active.

alifbe (approved):
LGTM 👍

PR #553, opened 2025-09-16
Fix date check: compare chosen dates to requested dates

michael-liem-4:
The length of chosendates should be compared against the number of dates requested by the user instead of the number of available dates in the UNRST file

### DIFF
--- a/tests/test_grid.py
+++ b/tests/test_grid.py
@@ -604,7 +604,7 @@ def test_get_available_rst_dates():
 
 
 def test_dates2rstindices_logging(caplog):
-    """Test warning is logged in dates2rstindices if not all requested dates are available."""
+    """Test warning is logged in dates2rstindices if not all dates are available."""
     resdatafiles = ResdataFiles(REEK)
     alldates = grid.get_available_rst_dates(resdatafiles)
 


michael-liem-4:
I came up with this test to assert this behaviour. The test uses pytest’s `caplog` fixture by declaring it as a function argument.

Do you prefer that I add it as a separate test in `tests/test_grid.py`, or should I rather adapt the existing test to `def test_get_available_rst_dates(caplog)` and add it there?

```python
import datetime
import logging

from res2df import ResdataFiles, grid

def test_dates2rstindices_warning(caplog):
    resdatafiles = ResdataFiles(REEK)
    alldates = grid.get_available_rst_dates(resdatafiles)

    # Case 1: check that a warning is logged if not all requested dates are available
    caplog.clear()
    with caplog.at_level(logging.WARNING):
        grid.dates2rstindices(resdatafiles, alldates + [datetime.date(1900, 1, 1)])
    assert "Not all dates found in UNRST" in caplog.text

    # Case 2: check that no warning is logged if all requested dates are available
    caplog.clear()
    with caplog.at_level(logging.WARNING):
        grid.dates2rstindices(resdatafiles, alldates[0])
        grid.dates2rstindices(resdatafiles, alldates[0:2])
        grid.dates2rstindices(resdatafiles, alldates)
    assert "Not all dates found in UNRST" not in caplog.text
```

PR #554, opened 2025-09-17
Add init file to opm keyword directory

larsevj:
~This avoids the opmkeyword folder being treated as a namespace package which does not mix very well with komodoenv and editable installs...~

Issue still persists, but I still think this is more correct.


berland:
> * This does currently not work since the paths have changed in opm-common, should this be fixed? 

yes!

PR #562, opened 2025-10-09
Add wildcard expansion of wlist wells in compdat

larsevj:


larsevj:
In `test_wlist.py` there is the following test case:
```
        # Wildcard wells should pass through, the dataframe user
        # will have to process it.
        (
            pd.DataFrame(
                [
                    {
                        "NAME": "OP",
                        "ACTION": "NEW",
                        "WELLS": "PROD*",
                        "DATE": datetime.date(1900, 1, 1),
                    },
                ]
            ),
            pd.DataFrame(
                [
                    {
                        "NAME": "OP",
                        "ACTION": "NEW",
                        "WELLS": "PROD*",
                        "DATE": datetime.date(1900, 1, 1),
                    },
                ]
            ),
        ),
    ],
```

It is unclear to me if this was due to the functionality not being in place at the time, for instance via the `get_wells_matching_template` function added in https://github.com/equinor/res2df/pull/333 or if there are any specific reasons or pitfalls related to supporting this functionality.
Do you remember anything @berland ?

larsevj:
> [89d14d7](https://github.com/equinor/res2df/commit/89d14d754c259b48beb05c776bff3befd3feb7c2) indicates that there might be a reason not to process wildcards in WLIST parsing.

Do you know what the commit message `Can be handled in WELOPEN instead.`, refers to? As I can currently understand, you end up inserting from wlist_df into welopen_df, and afterwards you try and match up the wells in `welopen_df` and `compdat_df`, but since they can be wildcarded in `wlist_df -> welopen_df` you run the risk of them not matching up with `compdat_df` as seen in https://equinor.slack.com/archives/CMS2PTC3Z/p1759499665007219

But there is an expansion of wildcards in welopen_df, only it happens before wells from wlist_df is inserted into welopen_df. So maybe if you do the wildcards expansion of welopen afterwards it would work, but then I do not see why you would not do it directly in wlist_df. Probably something I am missing here.

PR #584, opened 2026-01-22
Code cleanup, bug fixes, and test improvements

larsevj:


src/res2df/vfp/_vfp.py
```diff
@@ -188,7 +188,6 @@ def basic_data2pyarrow(data: dict[str, Any], /) -> pa.Table:
                 tab_type=data["TAB_TYPE"],
                 flow_values=data["FLOW_VALUES"],
                 thp_values=data["THP_VALUES"],
-                thp_indices=data["THP_INDICES"],
```
alifbe, line 191: Just curious, what is the reason we are dropping thp_indices?

src/res2df/common.py
```diff
@@ -977,4 +981,4 @@ def get_wells_matching_template(template: str, wells: list[str]) -> list[str]:
         )
     template = template.removeprefix("\\")
     regex = template.replace("*", ".*").replace("?", ".")
-    return [well for well in wells if bool(re.match(regex, well))]
+    return [well for well in wells if bool(re.fullmatch(regex, well))]
```
alifbe, line 980: ~~Do we have any tests that document this change?~~

Oops, see it down the line.

src/res2df/summary.py
```diff
@@ -831,12 +833,10 @@ def summary_reverse_main(args: argparse.Namespace) -> None:
     eclbase = Path(args.output).name
 
     # Summary.fwrite() can only write to current directory:
-    cwd = Path.cwd()
     summary = df2ressum(summary_df, eclbase)
-    try:
-        os.chdir(outputdir)
+    outputdir.mkdir(parents=True, exist_ok=True)
+
+    with contextlib.chdir(outputdir):
```
alifbe: I'm fully supportive for this, but I know that there was discussion that in FMU, all folder need to be created in advance.

tests/test_common.py
```diff
@@ -264,8 +263,9 @@ def test_eclcompress():
     "template, wells, output",
     [
         ("OP*", ["OP1", "OP2", "WI"], ["OP1", "OP2"]),
-        ("B*H", ["B_1H", "BH", "B_23H", "WI"], ["B_1H", "BH", "B_23H"]),
+        ("B*H", ["B_1H", "BH", "B_23H", "WI", "B_1HX"], ["B_1H", "BH", "B_23H"]),
         ("B_1H*", ["B_1H", "B_1HT2", "OB_1H"], ["B_1H", "B_1HT2"]),
+        ("OP?", ["OP1", "OP12", "OP"], ["OP1"]),
```
alifbe, line 268: I am a bit concern with this. I think the changes to fullmatch make sense and more accurate. But I see that this will a breaking change.

tests/test_wellconnstatus.py
```diff
@@ -96,6 +96,54 @@ def test_eightcells_dataset():
...
+                columns=["DATE", "CPI:OP1:1,1,1", "XCPI:OP1:2,2,2"],  # X at start
+            ),
+            pd.DataFrame(
+                [
+                    ["2000-01-02", "OP1", 1, 1, 1, "OPEN"],
+                ],
+                columns=["DATE", "WELL", "I", "J", "K", "OP/SH"],
+            ),
+        ),
+        # Well name too long (>8 chars) should be ignored
```
alifbe, line 131: I'm not so sure about this. I heard FLOW support well name more than 8 characters. But I am not sure if this will be stored in summary file as well.

src/res2df/vfp/_vfp.py
larsevj: They were not used at all within the function, and since i do not expect any users to access this function directly I thought it was clearer to drop them.

src/res2df/summary.py
larsevj: I see, will remove the `mkdir` call then

tests/test_wellconnstatus.py
larsevj: This should only demonstrate the current behaviour. The change from match to fullmatch in wellconstatus is merely for clarity. The re.match call had anchors ^ and $, so the behaviour should be equivalent. Added the tests before making the change to ensure they still passed.

src/res2df/summary.py
alifbe: Like I say, I am in favor of preparing the output folder instead of crashing. But perhaps ask around before decide on something.

Example in subscript https://github.com/equinor/subscript/blob/4ac49d50d4c853e12d987f4b7e7844fab697b621/src/subscript/prtvol2csv/prtvol2csv.py#L97-L102

Perhaps @berland has comment about this?

tests/test_common.py
larsevj: In the case of templates ending with a literal character it will no longer match wells that have other trailing characters. So `B*H` used to match `B_1HX`, which I do not hope anyone using it expected.

src/res2df/summary.py
berland: I support not making the directory implicitly.

src/res2df/parameters.py
```diff
@@ -125,49 +125,50 @@ def load(filename: str | Path) -> dict[str, Any]:
     """
     params_dict = None
 
-    if not Path(filename).exists():
+    if not Path(filename).is_file():
         raise FileNotFoundError(str(filename) + " not found")
-
-    if not Path(filename).read_text(encoding="utf-8").strip():
+    file_content = Path(filename).read_text(encoding="utf-8")
+    if not file_content.strip():
```
alifbe, line 131: [nitpick] Perhaps file_content can be stripped as well.

alifbe (approved):
Looks good to me 👍

PR #586, opened 2026-03-26
Use correct dtype in fillna for rft

larsevj:
User report about mixed types in HOSTNAME column causing failure when trying to convert to pyarrow later on. Ensure we fill in nan values with empty strings instead of a 0 (int).


berland (approved):
lgtm!

PR #597, opened 2026-07-27
fix: remove numpy upper bound and fix datetime comparison in test

### LINKED ISSUE #594: Resolve `numpy 2.5.1` PyArrow timestamp raising OverflowError
Starting with numpy 2.5.1, converting a PyArrow timestamp array containing such extreme dates back into a numpy datetime64 array (via np.array(...) ) now raises OverflowError: Overflow when converting between datetime64 units, where numpy 2.4.6 handled it without error.

```
___________________________ test_df2pyarrow_500years ___________________________

    def test_df2pyarrow_500years():
        """Summary files can have DATE columns with timespans outside the
        Pandas dataframe nanosecond limitation. This should not present
        a problem to the PyArrow conversion"""
        dateindex = [dt(1000, 1, 1, 0, 0, 0), dt(3000, 1, 1, 0, 0, 0)]
        dframe = pd.DataFrame(
            columns=["FOO", "BAR"], index=dateindex, data=[[1, 2], [3, 4]]
        ).astype("int32")
    
        # The index name should be ignored:
        dframe.index.name = "BOGUS"
        pyat = _df2pyarrow(dframe)
    
        assert (
>           np.array(pyat.column(0))
            == [
                np.datetime64("1000-01-01T00:00:00.000000000"),
                np.datetime64("3000-01-01T00:00:00.000000000"),
            ]
        ).all()
E       OverflowError: Overflow when converting between datetime64 units

/home/runner/work/res-compat/res-compat/tests/test_summary.py:967: OverflowError
```

rnyb:
- Remove numpy<2.5 version constraint from dependencies
- Replace numpy datetime64 array comparison with pylist check

Closes #594

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,7 +38,7 @@ dependencies= [
 "resdata>=5",
 "resfo",
 "networkx",
-"numpy<2.5",
+"numpy",
 "opm>=2020.10.2",
 "pandas >= 2",
 "pyarrow",
--- a/tests/test_summary.py
+++ b/tests/test_summary.py
@@ -1205,13 +1205,10 @@ def test_df2pyarrow_500years():
     dframe.index.name = "BOGUS"
     pyat = _df2pyarrow(dframe)
 
-    assert (
-        np.array(pyat.column(0))
-        == [
-            np.datetime64("1000-01-01T00:00:00.000000000"),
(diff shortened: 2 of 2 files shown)


alifbe (approved):
LGTM 👍

